PR #11, opened 2015-07-19
Google Maps integration [do not merge]

alvinwan:
Current target behavior; feel free to comment below with objections/suggestions/comments.

Steps:
- [ ] 1. develop a simple, working JS script
- [ ] 2. develop Python wrapper for JS script
- [ ] 3. integrate Python wrapper with IPython

## Notebook

Using [IPython magic](http://ipython.org/ipython-doc/dev/config/extensions/index.html), load into a notebook:

```
load_ext datascience.polymap
```

To plot, use `polymap` magic

```
polymap  [flags[args]]  # line magic, exists only in this line
%%polymap  [flags[args]]  # cell magic, persists... across cells?
```

A sample invocation:
```
%%polymap
outline Los Angeles, CA
```

Options:
- [ ] `outline` -- using a polygon, outlines an area; accepts anything Google Maps can recognize
- [ ] `label` -- draws a dot and labels an area; accepts anything Google Maps can recognize
- [ ] `center` -- lat-long pair at the center of the map
- [ ] `zoom` -- zoom level

## Python Development

A Python script will generate JS to then output for the outside world.

**Polymap** -- main wrapper for a map
- [ ] `label_map` -- adds all MapPoints and labels to the map
- [ ] `outline_map` -- adds all MapRegions to the map
- [ ] `draw_map` -- draws a map using MapRegion and/or MapPoints
- [ ] `output` -- generate javascript

**MapRegion** -- a single polygon

**MapPoint** -- a single point, with a label and potentially display information

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,7 @@ docs/_build/
 
 # PyBuilder
 target/
+
+# Idea Projects
+datascience.iml
+.idea/*
--- a/polymaps/__init__.py
+++ b/polymaps/__init__.py
@@ -1,0 +1,14 @@
+from .main import Polymap
+from IPython.display import Javascript, display
+
+
+def load_ipython_extension(ipython):
+	# The `ipython` argument is the currently active `InteractiveShell`
+	# instance, which can be used in any way. This allows you to register
+	# new magics or aliases, for example.
+	ipython.register_magics(Polymap)
+	
+	# Loads the JS extension.
+	display(Javascript("""
+		IPython.load_extensions('datascience_polymaps_js/polymaps_views')
+		"""))
--- a/polymaps/main.py
+++ b/polymaps/main.py
@@ -1,0 +1,21 @@
+# This code can be put in any Python module, it does not require IPython
+# itself to be running already.  It only creates the magics subclass but
+# doesn't instantiate it yet.
+from __future__ import print_function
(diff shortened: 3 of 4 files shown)


papajohn:
Can you check in `datascience_polymaps_js/polymaps_views`

alvinwan:
Eh, I was trying to emulate `gmap`. Not exactly sure how he was able to define that in the Python extension namespace.

We can take a different approach instead:

- compact the IPython magic script in one file and (for now) install it like this:
```
%install_ext https://raw.github.com/path/to/script.py
```
(ref. IPython [extensions index](https://github.com/ipython/ipython/wiki/Extensions-Index) and [a script that we can model ours after](https://github.com/debjan/ipython-recipes-magic/blob/master/recipes.py))
- preload our custom JS script with `jupyter-notebook-student` (for testing purposes, simply symlink the script to your IPython `notebook/statics/custom/` directory and add `custom/[script name]` to `main.js`)
- have the magic print HTML output, which will then use our custom, preloaded script

papajohn:
If possible, I like the loading scheme of

```
pip install datascience
%load_ext polymaps
```

rather than having multiple download points or hard-coding the integration. Is that possible?

alvinwan:
It should be possible; I'll have to investigate more.

PR #12, opened 2015-07-20
Simplify transform and read-only methods for tables

papajohn:


papajohn:
Addresses #10

alvinwan:
LGTM. Merging.

Issue #5, opened 2015-07-17
[moved] Draw polygons on a map from a notebook


alvinwan:
Prof. DeNero

> I think we want something like https://github.com/pbugnion/gmaps to draw a Google map with polygons & dots on it using the Google Maps Javascript API. It's unclear whether this library has been ported to Python 3. They have a directory that gives me hope: https://github.com/pbugnion/gmaps/tree/master/examples/ipy3

SamLau95:
The gmaps package was written in Python 2 so imports were broken all over the place when installing using Python 3. I made a fork that seems like it works okay at https://github.com/SamLau95/gmaps

SamLau95:
From what I can tell, the gmaps package doesn't have too much power. In fact, it pretty can only make heatmaps. But maybe these are good enough for what we're looking to do.

https://dl.dropboxusercontent.com/content_link/J6CIxgKTWer5npvahItOtJKuocUk5K5hnk8oBcEN6lEwM8wLpqSTRHgloB4muiGx

alvinwan:
Judging by project 1's title, it seems like we need the polygon to wrap around an area.

gmaps is pretty simple, but it seems like a good guide so far.

alvinwan:
Didn't know about the 2to3 tool either. That's pretty neat, what you did there.

papajohn:
We might have to write our own Python wrapper for the Google Maps
Javascript API that works in the IPython environment. I think we need
points and polygons more than heat maps. Hopefully gmaps can get us started
quickly.

deculler:
That's pretty much where I came to as well.  Ought to be able to use google
maps as a background image for basemap in matplotlib.

David E. Culler
Friesen Professor of Computer Science
Electrical Engineering and Computer Sciences
University of California, Berkeley

papajohn:
I have an hour to kill. Any suggestions about how I can make progress on this issue? I'll start by reading the Google Maps Javascript API.

SamLau95:
I'm going to be working on this later tonight. I think that's a good place to start; any documentation / recommendations on that would save us time searching.

papajohn:
I'm sending everyone an API key over email, since this repo is public. Feel
free to use it as much as you need.

The Maps Embed API is not sufficient for our needs. We have to use the
Javascript API.

alvinwan:
@papajohn @SamLau95 

Here's how I broke down the application.

IPython integration:
[...]

PR #17, opened 2015-07-22
Add folium integration and use it in an example notebook

SamLau95:
With the discovery of the [folium](https://github.com/python-visualization/folium) package, I remove the old hacky stuff
I put in before for gmaps and use folium instead!

I also added an example notebook in the `examples/` folder that
successfully renders a map centered at Broad Street (now Broadwick
Street).

The next step is to add the points / polygon functionality, and we can call #5 a wrap.


SamLau95:
screenshot 2015-07-21 23 00 23

datascience/maps.py
```diff
@@ -19,8 +14,27 @@ def draw_map(center, zoom, points=[], regions=[]):
     zoom -- zoom level
     points -- a sequence of MapPoints
     regions -- a sequence of MapRegions
+
+    TODO(sam): Add points and regions functionality
     """
-    # Some magic to get javascript to display the map
+    map = folium.Map(
+        location=center,
+        tiles='Stamen Toner',
```
SamLau95: Note that I chose this arbitrarily — if you feel like there's a better map style to use let me know.

SamLau95:
Also — I didn't increment the package version. Do we increment the version on every PR?

alvinwan:
That's fine - we only increment when we republish the package.

This feature would definitely warrant a new version, but we don't know how many other deployments we might make before this is ready.

PR #18, opened 2015-07-23
Add ability to draw points

SamLau95:
Github is unfortunately bad at PRs that depend on other PRs. See https://github.com/dsten/datascience/compare/setup-folium...draw-points for the actual changes made between #17 and this PR.

This is essentially a wrapper around the [`circle_marker` method](https://github.com/python-visualization/folium/blob/master/folium/folium.py#L458) from
`folium` and keeps the same arguments.

Now we can do:

```python

>>> points = [MapPoint((51.5135015, -0.1358392))]
>>> draw_map((51.5135015, -0.1362392), points=points)
<hooray it's a map!>

```

Now all that's left to finish #5 is polygons which seem a bit more
involved since the folium primitive isn't what we want.

Depends on #17


SamLau95:
screenshot 2015-07-22 23 25 30

SamLau95:
Github is unfortunately bad at PRs that depend on other PRs. See https://github.com/dsten/datascience/compare/setup-folium...draw-points for the actual changes made between #17 and this PR.

PR #19, opened 2015-07-23
Add ability to draw polygon

alvinwan:
builds off of #18 

I thought I'd play with it.

Summary:
- [x] using @SamLau95's efficient structure and approach, wrapping
- [x] can now pass list of lat-long pairs, list of list of lat-long pairs, or list of MapRegions to MapRegion
- [x] autozoom based on given coordinates (each region could have a suggested "frame" and parents somehow aggregate suggested frames)
- [x] object wrapper for geoJSON files

**Current Syntax and Capabilities**

```
# native folium functionality
region = MapRegion(geo_path='../data/us-states.json')
draw_map((51.5135015, -0.1362392), regions=[region], zoom=3)

# using Data, simplify syntax, and use data selectively
states = Data('../data/us-states.json')
draw_map(states.CA)

states_list = states.to_list()
draw_map(regions=states_list[3:5])

states_data = states.to_dict()
draw_map(regions=list(states_data.values()))
include = lambda state: state.startswith('A') and state.endswith('K')
draw_map(regions=list(filter(include, states_data.items()))

# points are just a list of lat-long pairs
points = [(51.5135010, -0.1358392), (51.5135015, -0.1358392), (51.5135010, -0.1358398)]
region1 = MapRegion(points=points)
map_html = draw_map((51.5135015, -0.1362392), regions=[region1, region2])

# For regions with multiple polygons (i.e., Alaska), use list of lists, or list of list of lists, or list of list of....
locations = [[(51.5135015, -0.1358392), (51.5135020, -0.1358392), (51.5135015, -0.1358398)], 
			[(51.5135020, -0.1358392), (51.5135025, -0.1358392), (51.5135020, -0.1358398)]]
region2 = MapRegion(locations=locations)

# a MapRegion may take MapRegions recursively
papaRegion = MapRegion(regions=[region1, region2])
grandpaRegion = MapRegion(regions=[papaRegion])
draw_map((51.5135015, -0.1362392), regions=[grandpaRegion])  # draw with grandpa attributes
draw_map((51.5135015, -0.1362392), regions=[papaRegion])  # draw with papa attributes

# control how children MapRegions are rendered
# achieve this effect: draw_map(regions=[region1, region2])
grandpaRegion.composite = MapRegion.CHILDREN
draw_map(regions=[papaRegion])

# create either a MapPoint or MapRegion by passing it as the first arg, let autozoom and auto center take care of the rest
draw_map(grandpaRegion)
```

**Demonstration**

maps-complete


alvinwan:
MapRegions can now accept:
- all of the original `folium` params (a `geo_json` source file)
- list of lat-long pairs
- list of list of lat-long pairs
- list of MapRegions (and they can be defined recursively... infinitely)

The default behavior for MapRegions is relatively efficient; it simply combines all region data into one massive JSON, *then* renders that JSON. Behavior can be changed, though:
- `MapRegion.GROUP` -- default, renders direct descendants with this MapRegion's attributes
- `MapRegion.CHILDREN` -- renders direct descendants with each descendants' own attributes (inefficient as of this moment, invokes `map()` for each descendant-region)
- `MapRegion.ONE` -- mushes together *all* descendants into one multi-polygon, renders with parent attributes (for now)

tests/test_map.py
```diff
@@ -0,0 +1,113 @@
...
+	
+	grab_locations = lambda json_data: json_data['features'][0]['geometry']['coordinates']
+
+	json_data = region1._to_json()
+	locations = grab_locations(json_data)
+	assert locations == locations1
+
+	json_data = region2._to_json()
+	locations = grab_locations(json_data)
+	assert locations == [locations2]
```
SamLau95: testing ++

datascience/maps.py
```diff
@@ -0,0 +1,346 @@
...
+        # self.attributes['geo_str'] = json_str
+        if self['locations'][0] or self['points'] or self['regions']:  # TODO: temporary workaround
+            import os
+            try:
+                os.mkdir('cache/')
+            except FileExistsError:
+                pass
+            self.TEMP_FILE = 'cache/'+str(round(random.random()*100))+'temp.json'
+            open(self.TEMP_FILE, 'w').write(json_str)
+            self.attributes['geo_path'] = self.TEMP_FILE
```
alvinwan: @SamLau95 I couldn't get `geo_str` to work; do you know what it's supposed to take?

I know the `geoJSON` is valid, so I employed an ugly hack... for now - wrote it to a file and had folium read from the file.

datascience/maps.py
```diff
@@ -0,0 +1,510 @@
...
+#########################
+# Exposed Functionality #
+#########################
+
+
+def draw_map(center=None, **kwargs):
+    """ 
+    Draw a map with center & zoom containing all points and
+    regions that displays points as circles and regions as polygons.
+    """
```
SamLau95: For someone who's using `help(draw_map)`, the info that'll be displayed here isn't super helpful. What can I pass in to `**kwargs`? Just looking at the method signature itself and the docstring doesn't tell the user what to call `draw_map` with, so we should list out the argument list explicitly either in the method sig itself or in the docstring.

(This is one of the reasons why I somewhat prefer explicitly listing the arguments; it's more verbose but also more helpful to someone looking at it for the first time.)

datascience/maps.py
```diff
@@ -0,0 +1,510 @@
...
+        k: v for k, v in rval.items() 
+        if k not in excludes
+        and k in (includes or rval.keys())
+    }
+
+
+def _hook(f):
+    """
+    Decorator for obj method, checks for PREFIX_before and PREFIX_after 
+    a method is called
```
SamLau95: nit: This doesn't just check for `PREFIX_before / after`, it calls 'em, so we can be more clear here.

datascience/maps.py
```diff
@@ -0,0 +1,510 @@
...
+######################
+# Folium Abstraction #
+######################
+
+
+class MapEntity:
+    """
+    General Map entity
+    - attributes get and set, using map[prop]
+    - all other data get and set, using map.prop
```
SamLau95: I think this could be more clear as well. When I read `General Map entity` I thought this was a abstract class that only different types of `Map` inherit from. But, it looks like all our other map classes inherit from this one. So this is more like a class that helps with data getting / setting? I'm not entirely sure. I'd love to know why this class is here, as I'm sure there's a good reason but it's not immediately obvious from the docstring.

What's the difference between an attribute and `other data`?

datascience/maps.py
```diff
@@ -0,0 +1,510 @@
...
+        #     self._autofit(self.bounds)
+    
+    def _before_map(self, *args, **kwargs):
+        """ Autozoom map where appropriate """
+        self.bounds = self._autobounds()
+        
+        if not self.location:
+            self._autocenter(self.bounds)
+
+        # remove with new Folium release
```
SamLau95: nit: Maybe you can leave your name here so a future dev knows who to ask about this? I leave comments like `TODO(sam): do stuff` for this purpose.

datascience/maps.py
```diff
@@ -0,0 +1,510 @@
...
+            width = ((bounds['max_lat'] - bounds['min_lat']) or 2)
+            height = ((bounds['max_long'] - bounds['min_long']) or 2)
+            area, max_area = width*height, 180*360
+            zoom = math.log(area/max_area)/-factor
+            self['zoom_start'] = max(1, min(18, round(zoom)))
+        except ValueError:
+            raise MapError('Uh oh. Is the data formatted as long-lat \
+            pairs? (Check. If the smallest number is less than -90, \
+            those are long-lat pairs) If so, please add geo_formatted=True \
+            to create_map')
```
SamLau95: This is a very helpful error message :)

SamLau95:
nit: I noticed a number of trailing whitespace characters and files without newlines at the end. You can fix this by adding

```
"ensure_newline_at_eof_on_save": true,
"trim_trailing_white_space_on_save": true
```
to your user preferences in Sublime (I'm sure vim and emacs have ways of doing this as well).

The code in general looks good! There is definitely a lot going on in this PR, so for someone with a very size-limited stack in my head it's difficult to review. I think each of your checkbox tasks could be its own PR.

alvinwan:
Thanks for the code review -- was just about to ask for one. I can make corrections tomorrow.

Thanks again for the Sublime suggestion.

datascience/maps.py
alvinwan: Oh, I see. Never knew about `help`.

datascience/maps.py
alvinwan: Mhm, yes. To some degree, it's a utility for getting/setting. It's mostly to help make all the entities uniform, so that calling `map()` on anything, will map it.

An attribute is something that'll be passed to folium, in the class's `map()` method. `other data` is any data you'd like the object to have.

I'll clarify the docstrings.

datascience/maps.py
alvinwan: Ah, I see. Added.

alvinwan:
@deculler @papajohn Would one or both of you like to make one more pass before I merge and package a new release?

@SamLau95 I'll break up PRs in the future as well. Thanks.

deculler:
Sure.  I will build some examples over the weekend.  Seems like we might
want to make it jive better with tables.

datascience/maps.py
alvinwan: :D Thanks

deculler: Cool.  Seems to work at reasonable scale.

papajohn:
Looks great from what I've read so far! I need to keep reading for a bit (probably tomorrow), but I'll merge very soon. 

I am going to move all of the data and examples to the `text` repo, leaving `datascience` with just code.

alvinwan:
Awesome! Sounds good to me.

Issue #24, opened 2015-07-28
Table help is incorrect


davidwagner:
`help(Table)` gives the following example of how to use a `Table`:

     |  >>> letters = ['a', 'b', 'c', 'z']
     |  >>> counts = [9, 3, 3, 1]
     |  >>> points = [1, 2, 2, 10]
     |  >>> t = Table([('letter', letters), ('count', counts), ('points', points)])

However copy-pasting that into Jupyter goes the following error message:

    ---------------------------------------------------------------------------
    AssertionError                            Traceback (most recent call last)
    ...
    /usr/local/lib/python3.4/dist-packages/datascience/table.py in __init__(self, columns, labels)
    ---> 46         assert labels is not None, 'Labels are required'

alvinwan:
Sorry, table syntax changed, and I never updated it. Fixed:

```
letters = ['a', 'b', 'c', 'z']
counts = [9, 3, 3, 1]
points = [1, 2, 2, 10]
t = Table([letters, counts, points], ['letter', 'count', 'points'])
print(t)
```

Issue #22, opened 2015-07-28
Join tests broken


alvinwan:

deculler:
This may be a result of going to left outer join.  If you look at the
berkeley_tables in demo you'll see why it is so nice to have.  For the case
where there is a unique item in each table, these all are the same.  I do
think the asymmetry is quite natural.

David E. Culler
Friesen Professor of Computer Science
Electrical Engineering and Computer Sciences
University of California, Berkeley

alvinwan:
@deculler I see I see. However, it looks like the left outer join isn't preserving unmatched rows from the "left" (first) table.

Output of table.join('points', table2)
```
E      points | letter | count | totals | names
E      2      | c      | 3     | 6      | two
E      2      | b      | 3     | 6      | two
E      1      | a      | 9     | 9      | one
```

where table is:
```
	letter | count | points
	a      | 9     | 1
	b      | 3     | 2
	c      | 3     | 2
	z      | 1     | 10
```

From `table`, the last row is missing. Right now, it looks like it's still an inner join, except with duplicate matches in the indexed column, repeated.

Let me know if I'm missing something, or if I should fix it.

alvinwan:
I went ahead and changed `join` so that it included rows in left and not in right, amended tests.

deculler:
The result only includes matched rows.  The difference we were after is
that a row in the right might match multiple rows in the left.  There would
be a row in the result for each left row that had a matching right.  The
more general form would have a row in the result for each pairwise match.

Thanks for looking into this.

David E. Culler
Friesen Professor of Computer Science
Electrical Engineering and Computer Sciences
University of California, Berkeley

alvinwan:
Ah, I see. Sounds good.

Yep.

Issue #2, opened 2015-07-17
[moved] Appending a table is inefficient


alvinwan:
Prof. DeNero:

> `np.append` should be invoked once per column.

PR #26, opened 2015-07-29
Table fixes + optimizations

alvinwan:
- [x] #2 `append` invoked only once per column when append table to table
- [x] #4 can append two tables with columns in different order (second table can have extra columns, but those will be ignored)
- [x] #22 fix join and join tests
- [x] #23 fixes sort tests
- [x] #24 fixes docstring for `Table`


alvinwan:
@papajohn @deculler Ready for review.

PR #33, opened 2015-07-31
Sample with or with_replacement from rows, with weights if desired

alvinwan:
depends on #32 

```
sample(k, with_replacement=False, weights=None)
```

After toying around for a bit, found `numpy.random.choice`

### DIFF
--- a/datascience/table.py
+++ b/datascience/table.py
@@ -548,14 +548,20 @@ class Table(collections.abc.Mapping):
             index.setdefault(key, []).append(row)
         return index
     
-    def _sample(self, k, with_replacement=False):
+    def _sample(self, k, with_replacement, weights):
         """Returns list of sampled rows"""
-        return random.sample(self.rows, k)
+        if weights:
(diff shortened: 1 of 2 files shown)


alvinwan:
@mijordan3 Is there anything else we need `sample` to do? I'm currently using a built-in numpy method. If we want more, I'd probably have to start from scratch.

mijordan3:
Hi Sam,

You've been working on some code for resampling-based hypothesis testing?
Can you coordinate with
Alvin, who's worked on a "sample" function that takes a table into a
randomly sampled table, so that we
have an integrated set of code for this functionality?

It would be good to have both bootstrap-based hypothesis testing (where
column A is resampled n times
with replacement and column B is independently sampled n times with
replacement), and permutation
testing (where column A and column B are put together in a single long
column, a random permutation
is made of that column, it is split in the middle into two new columns, and
the statistic is computed on
those two columns).

Thanks,
Mike

Issue #31, opened 2015-07-29
Pivot needs to support multiple rows


papajohn:
I already ran into an example where I needed multiple rows in a pivot. I had to join and then split a column to get what I wanted, which was gross. We need to support something like

`t.pivot(column, [row0, row1], value, ...)`

mDibyo:
Working on this. Not able to/couldn't figure out how to assign it to myself...

alvinwan:
@papajohn has to add you to the org. Couldn't find you in the assignee list on the right.

mDibyo:
Added to the organization. Thanks Alvin!

PR #35, opened 2015-08-04
Fix Issue #31: Pivot needs to support multiple rows

mDibyo:
Travis build seems to be failing... Any idea why?

Otherwise, ready for review.


mDibyo:
Adding link to Issue #31...

SamLau95:
Looks like travis is failing cause it isn't configured yet (it's trying to run `rake` instead of `python setup.py test`). @alvinwan should be taking care of that soon.

datascience/table.py
```diff
@@ -320,22 +320,30 @@ def groups(self, column_labels, collect=lambda s: s):
...
 
         columns, rows, values -- column labels in self
         collect -- aggregation function over values
         zero -- zero value for non-existent row-column combinations
         """
-        selected = self.select([columns, rows, values])
-        grouped = selected.groups([columns, rows], collect)
-        row_values = np.unique(self._get_column(rows))
-        pivoted = Table([row_values], [rows])
+        if not _is_non_string_iterable(rows):
```
papajohn: `rows = _as_labels(rows)` should suffice.

tests/test_table.py
```diff
@@ -253,6 +253,22 @@ def test_pivot(t):
...
+	t['early'] = t['letter'] < 'd'
+	t['late'] = t['letter'] > 'c'
+	t['exists'] = 1
+	test = t.pivot('points', ['early', 'late'], 'exists')
+	assert_equal(test, """\
+	early | late  | 1 exists | 2 exists | 10 exists
+	False | False | None     | None     | None
+	False | True  | [1]      | None     | [1]
+	True  | False | [1]      | [1 1]    | None
+	True  | True  | None     | None     | None
```
papajohn: This row shouldn't exist in the output.

datascience/table.py
```diff
@@ -320,22 +320,30 @@ def groups(self, column_labels, collect=lambda s: s):
...
+            rows = [rows]
+        selected = self.select([columns, values] + rows)
+        grouped = selected.groups([columns] + rows, collect)
+
+        # Generate all possible combinations of values from columns in rows
+        rows_values = \
+            list(itertools.product(*(np.unique(self._get_column(column_label))
```
papajohn: I think we actually want only sets for which there exists at least one row, rather than the `product` of all possible combos. So, `np.unique` the tuples of row values.

datascience/table.py
```diff
@@ -543,9 +551,9 @@ def index_by(self, column_or_label):
         rows corresponding to each value.
         """
         column = self._get_column(column_or_label)
-        index = {}
+        index = collections.defaultdict(list)
         for key, row in zip(column, self.rows):
-            index.setdefault(key, []).append(row)
+            index[key].append(row)
         return index
```
papajohn: `return dict(index)` so that we don't leak `defaultdict`s out into the world (or just revert).

datascience/table.py
```diff
@@ -749,17 +757,22 @@ def wrapped(column):
     return wrapped
 
 
-def _fill_with_zeroes(order, rows, zero=None):
-    """Return a column of the index-1 elements in rows, where the index
-    of each value is determined by matching index-0 to an element of keys.
+def _fill_with_zeroes(keys, rows, zero=None):
+    """Return a column of the index-(-1) elements in rows, where the index
```
papajohn: How about `(index - 1)` to indicate it's an expression.

datascience/table.py
```diff
@@ -749,17 +757,22 @@ def wrapped(column):
     return wrapped
 
 
-def _fill_with_zeroes(order, rows, zero=None):
-    """Return a column of the index-1 elements in rows, where the index
-    of each value is determined by matching index-0 to an element of keys.
+def _fill_with_zeroes(keys, rows, zero=None):
```
papajohn: I vote for "zeros" over "zeroes"

datascience/table.py
papajohn: I have a really hard time figuring out what this function does. Now seems like a good time to clean up the description or reduce its complexity in some way. Like, what's `index`?

datascience/table.py
```diff
@@ -749,17 +757,22 @@ def wrapped(column):
...
+    """Return a column of the index-(-1) elements in rows, where the index
+    of each value is determined by matching index-(0:-1) to an element of
+    keys.
     """
     assert len(rows) > 0
-    index = dict(rows)
+    if not _is_non_string_iterable(keys):
+        keys = [(k,) for k in keys]
```
papajohn: I think everything is being wrapped in a tuple b/c you're assuming `rows` contains tuples and so slicing will give a tuple. Seems fine, but should be documented in the description.

datascience/table.py
```diff
@@ -320,22 +320,30 @@ def groups(self, column_labels, collect=lambda s: s):
...
+        # Generate all possible combinations of values from columns in rows
+        rows_values = \
+            list(itertools.product(*(np.unique(self._get_column(column_label))
+                                     for column_label in rows)))
+        pivoted = Table.from_rows(rows_values, rows)
+
+        # Generate other columns and add them to pivoted
         by_columns = grouped.index_by(columns)
         for label in sorted(by_columns):
             pairs = [t[1:] for t in by_columns[label]] # Discard column value
```
papajohn: These aren't necessarily pairs anymore, right? Let's update the names.

papajohn:
I added a few small comments, but looks like the right idea.

datascience/table.py
```diff
@@ -320,22 +320,28 @@ def groups(self, column_labels, collect=lambda s: s):
...
-        selected = self.select([columns, rows, values])
-        grouped = selected.groups([columns, rows], collect)
-        row_values = np.unique(self._get_column(rows))
-        pivoted = Table([row_values], [rows])
+        rows = _as_labels(rows)
+        selected = self.select([columns, values] + rows)
+        grouped = selected.groups([columns] + rows, collect)
+
+        # Generate existing combinations of values from columns in rows
+        rows_values = sorted(list(set(self.select(rows).rows)))
```
mDibyo: 'np.unique' was giving wrong results here (`list(np.unique([(True, True), (True, False)])) = [False, True]`).

datascience/table.py
```diff
@@ -749,17 +754,32 @@ def wrapped(column):
     return wrapped
 
 
-def _fill_with_zeroes(order, rows, zero=None):
-    """Return a column of the index-1 elements in rows, where the index
-    of each value is determined by matching index-0 to an element of keys.
+def _fill_with_zeros(partials, rows, zero=None):
```
mDibyo: Don't mind zeros.
How does this look?

mDibyo:
Made changes as suggested.

Also, I changed indentation in test_table and test_map from tabs to 4 spaces (PEP guidelines :D)

datascience/table.py
papajohn: Much clearer!

papajohn:
Looks good to me. Feel free to merge after resolving merge conflicts.

PR #40, opened 2015-08-05
Add percentile function to Table

### LINKED ISSUE #37: Implement a percentile function
The pth percentile of a list is the smallest number that is at least as large as p% of the numbers in the list.

That means: sort the list from low to high, go p% up the list from the bottom. If you're at a place on the list, take its value. Else take the next one.

SamLau95:
```
>>> print(table)
letter | count | points
a      | 9     | 1
b      | 3     | 2
c      | 3     | 2
z      | 1     | 10

>>> table.percentile('count', 70)
9
>>> table.percentile('count', 50)
3
```

Verified that `python setup.py test` works.
Resolves #37.

### DIFF
--- a/datascience/__init__.py
+++ b/datascience/__init__.py
@@ -1,3 +1,4 @@
 from .tables import *
 from .formats import *
 from .maps import *
+from .util import *
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -14,6 +14,7 @@ import IPython
 
 import datascience.maps as _maps
 import datascience.formats as _formats
+from .util import *
 
 class Table(collections.abc.Mapping):
     """A sequence of labeled columns.
@@ -445,8 +446,7 @@ class Table(collections.abc.Mapping):
         count | points
         9     | 10
         """
-        percentiles = [np.percentile(self[column_name], p, interpolation='higher')
-            for column_name in self]
+        percentiles = [percentile(self[column_name], p) for column_name in self]
         return Table(percentiles, self.column_labels)
     ##################
     # Export/Display #
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -427,6 +427,13 @@ class Table(collections.abc.Mapping):
             assert len(c) == self.num_rows, 'column length mismatch'
             return c
 
+    def percentile(self, label, p):
+        """Assumes that the column contains numbers.
+
+        Returns the smallest number that is at least as large as the p% of
+        numbers in the column.
+        """
+        return np.percentile(self[label], p, interpolation='higher')
     ##################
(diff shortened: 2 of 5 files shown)


papajohn:
Great! What do you think of making the interface like `sum`, in that it returns a one-row table with the stat computed for each column.

Also, we should add a higher-order `percentile` function that takes a percentage and returns a function on arrays. That way, we could pass it into `Table.stats`. E.g., `t.stats([np.sum, len, percentile(25)])`, perhaps.

```
>>> print(t)
a
1
2
3
>>> print(t.sum())
a
6
```

papajohn:
Actually, maybe it's not crazy to make the `percentile` function currying on one arg (a percent) but compute the percentile value on two args (a percent and an array).

SamLau95:
@papajohn could you clarify your last comment? Should I still add a `percentile` function at the top level?

Also, the build is failing because of `test_date_format` (https://travis-ci.org/dsten/datascience/builds/74514799), which doesn't seem related to my diff. `python setup.py test` passes on my machine.

papajohn:
The build failure is some version error we don't understand (or maybe time
stamps aren't consistent.

Yes, I think there should be a top level percentile function.

SamLau95:
I added a `util` module and implemented a `percentile` function there. One way to use it for `Table.stats` is to use `functools.partial`. Or, we could add a partial'd function from `util` as well. Any opinions?

papajohn:
I think we just write

def percentile(n, vs=None):
  if vs is None:
    return lambda vs: percentile(n, vs)
  # your implemention assuming vs is a sequence
  return some_number

Yes, functools.partial is the standard answer, but it's not a good fit for
this class.

-John

deculler:
Wouldn't it make sense to be able to pass in a vector of p's, in order, and
produce a row for each value of the other?

David E. Culler

SamLau95:
@papajohn Thoughts on David's suggestion? You have more context as to whether that functionality would be useful for students.

a-adhikari:
I like it. We'll need p = 0.25, 0.5, 0.75 for box plots anyway.
Ani

On Thu, Aug 20, 2015 at 12:03 AM, Sam Lau <notifications@github.com> wrote:

> @papajohn <https://github.com/papajohn> Thoughts on David's suggestion?
> You have more context as to whether that functionality would be useful for
> students.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dsten/datascience/pull/40#issuecomment-132913357>.
>



-- 
Ani Adhikari
Teaching Professor
Department of Statistics

papajohn:
That sounds reasonable to me.

PR #44, opened 2015-08-13
New maps interface

papajohn:
I simplified the maps interface and changed the code structure. There's less functionality now, but I think that's a good thing. All regions are completely atomic — can't inspect them except to see a bit list of points. Markers are just markers without a circle radius. 

Though I changed a lot, the original implementation was immensely helpful.

See https://github.com/dsten/text/pull/70 for demos of the new interface. Can you take a look and make suggestions?


papajohn:
Ok, I fixed a bunch of small issues. I think the interface is more stable now. I've made sure that @deculler 's Berkeley demo works again in the `map` branch of `text`.

tests/test_maps.py
```diff
@@ -40,177 +16,136 @@ def region2(points):
...
 
-def test_draw_marker():
-    """ Tests that draw_map returns HTML """
-    points = [MapPoint((51.5135015, -0.1358392))]
-    map_html = draw_map((51.5135015, -0.1362392), points=points)
-    assert isinstance(map_html, HTML)
 
+def test_marker_html():
+    """ Tests that a Marker can be rendered. """
+    Marker(51.514, -0.132).show()
```
alvinwan: Wow.

tests/test_maps.py
```diff
@@ -40,177 +16,136 @@ def region2(points):
...
+    Map([marker]).show()
 
-    point = MapPoint((51.5135015, -0.1358392))
-    map_html = draw_map(point)
-    assert isinstance(map_html, HTML)
 
+def test_map_region(states):
+    """ Tests that a Map can contain a Region. """
+    region = states['CA']
+    Map([region]).show()
```
alvinwan: This is really neat; could `Map` accept region/marker as a positional argument? i.e., wrap the `features` arg as a tuple if a subclass of `_FoliumWrapper`.

alvinwan:
Cool, the syntax is definitely much more coder-friendly. LGTM. Feel free to merge.

tests/test_maps.py
papajohn: Good idea. I'll do that.

Issue #50, opened 2015-09-02
Methods need better documentation


SamLau95:
When working with the datascience package, I spend a lot of time trying to figure out how the methods work since the docstrings aren't super helpful — some methods require the table to be a certain shape, others require the table values to be numbers, others strings. None of these details are mentioned in some important methods like `hist` and `barh`.

In addition, to find out whether the package has the functionality I want (eg. whether I can group a table of years by decade) I have to browse the methods one by one, trying to keep a lot of things in my head about what methods are available.

I imagine I'm running into a majority of these issues because most of this code wasn't written by me. However, this will be the case for our students so IMO the earlier we can work on this the better.

**It'd be very helpful to 1. Improve the docstrings and 2. Have easily navigable documentation (probably generated from docstrings using something like [Sphinx](http://sphinx-doc.org/)).**

A great place to start would be the plotting functions, since those seem to be the most finicky and most commonly used.

henryem:
The first lab on histograms will be next Wednesday (9/9), so this is somewhat urgent.  Even to design the lab over the next week, we'll need to understand what functionality is available.

papajohn:
Let's make progress! I've assigned Sam to get started. When you get stuck, ping @alvinwan and me. It's much easier to write effective docs if you're not the one who designed the code, I think.

PR #51, opened 2015-09-04
Add more documentation to hist and barh

SamLau95:
Also makes `Table.hist` and `Table.barh` throw a more helpful error message when the
Table contains invalid values (`np.flexible`, which includes strings).
Before, the method raised an error about not being able to operate on
"flexible types", which is not extremely helpful without knowledge of
numpy types.

Starts to address #50.

Travis tests are failing because of a long-standing issue with the date formatter.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -667,7 +667,35 @@ class Table(collections.abc.MutableMapping):
         return t
 
     def hist(self, overlay=False, **vargs):
-        """Draw histograms of all columns."""
+        """Requires all columns in the table to contain numerical values only.
+        If the columns contain other types, a ValueError is raised.
+
+        Draw one histogram per column. If the overlay argument is True, a legend
+        containing the column name is shown on each histogram.
+
+        See http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.hist
+        for additional arguments that can be passed into vargs. These include:
+            bins, range, normed, cumulative, and orientation, to name a few.
+
+        >>> table
+        count | points
+        9     | 1
+        3     | 2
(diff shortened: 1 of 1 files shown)


SamLau95:
@papajohn @alvinwan Either one of you wanna give this a review?

datascience/tables.py
```diff
@@ -582,10 +582,82 @@ def annotate(axis, ticks):
...
+        A total of n - 1 charts are created where n is the number of columns
+        in the table.
+
+        Note that the order of the categories in the table is not guaranteed to
+        be preserved in the bar graph. Ex. `barh` on a table with "a", "b", "c"
+        as the rows in the `column_for_categories` may not output a bar graph
+        with the labels in that order.
+
+        Requires every column except for `column_for_categories` to be
+        numberical. If the columns contain other types, a `ValueError` is
```
papajohn: numerical

datascience/tables.py
```diff
@@ -582,10 +582,82 @@ def annotate(axis, ticks):
...
+
+        Each chart is categorized using the values in `column_for_categories`
+        and one chart is produced for every other column in the table.
+        A total of n - 1 charts are created where n is the number of columns
+        in the table.
+
+        Note that the order of the categories in the table is not guaranteed to
+        be preserved in the bar graph. Ex. `barh` on a table with "a", "b", "c"
+        as the rows in the `column_for_categories` may not output a bar graph
+        with the labels in that order.
```
papajohn: Maybe we should just fix this behavior... filing an issue.

papajohn:
Yes, that's much improved!

PR #72, opened 2015-09-16
gitignore docs folder

SamLau95:
Because Travis autobuilds + pushes docs we can gitignore this so that
we don't have to have doc commits


SamLau95:
Just kidding, we just want to delete the build folders.

PR #73, opened 2015-09-16
gitignore built doc files

SamLau95:
Since Travis builds + pushes these automatically we can gitignore 'em so
that we won't have commits with built files.


SamLau95:
Nevermind, this causes Travis to not see the files to commit either :(

PR #76, opened 2015-09-23
Histogram from counts

papajohn:


datascience/tables.py
```diff
@@ -861,25 +877,36 @@ def hist(self, overlay=False, **vargs):
         <histogram of values in points>
         """
         # Check for non-numerical values and raise a ValueError if any found
-        # TODO(sam): Is a ValueError the right thing to raise?
```
SamLau95: Thanks for verifying this :)

PR #79, opened 2015-09-25
Add notes on Zenhub usage and workflow

SamLau95:
Hopefully this will help us manage and prioritize our issues better!
Later on we can add some more documentation on how to use it with
multiple repos.

See https://github.com/dsten/datascience/tree/zenhub-usage#using-zenhub for an easier-to-read version of the notes.

@papajohn Can you look it over and maybe even take a look at the board for this repo?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,50 @@ python setup.py test
 The recommended environment for installation and tests is the
 [Anaconda Python3 distribution](http://continuum.io/downloads#py34)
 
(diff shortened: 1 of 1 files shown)


papajohn:
Looks good to me, though I couldn't get boards to work for the private `text` repo.

SamLau95:
Right; I think you have to pay for it to work for private repos. However, I know that they provide this for free for nonprofits / educational orgs so I'll contact them.

PR #85, opened 2015-09-27
Tables & Histograms

papajohn:


datascience/tables.py
```diff
@@ -801,6 +801,100 @@ def annotate(axis, ticks):
             height = vargs.pop('height')
         self._visualize(labels, yticks, overlay, draw, annotate, height=height)
 
+    def bar(self, column_for_categories, overlay=False, **vargs):
```
papajohn: There must be a cleaner way to implement bar charts based on existing code for barh. I'll take a look.

datascience/tables.py
```diff
@@ -926,15 +1020,26 @@ def hist(self, overlay=False, bins=None, counts=None, **vargs):
         if overlay:
             # vargs.setdefault('histtype', 'stepfilled')
             plt.figure(figsize=(6, 4))
-            plt.hist(columns.values(), color=colors, **vargs)
+            (vals,rbins,patches) = plt.hist(list(columns.values()), color=colors, **vargs)
```
papajohn: How about we have separate methods for binning to create a table and drawing a histogram, rather than gluing them together like matplotlib. I think you usually either want to see the histogram or compute with it, but not very often both.

PR #92, opened 2015-10-02
Table.values and Table.append_column

### LINKED ISSUE #80: Let's phase out the usage of __getitem__ and __setitem__
The usage of these methods is a constant source of confusion for students, especially those that haven't programmed before. Many students are often confused about the usage of brackets versus parentheses. For example, whereas we first introduce bracket notation for lists and arrays using something like:

    my_array[0]
    my_array[2:5] # Slicing

For tables, all of a sudden they have to use strings inside of the brackets.

    my_table['Salary'] # Why?

In addition, we can't use slicing for obvious reasons to us but not to students.

**I think table usage would be much simpler and more consistent if the only way to access/modify values was through method calls.**

Students also mix up the usage of `__getitem__` and `select` — the first returns an array of values, the second returns a new table, and students basically have to memorize that fact. I think that if we taught them to use a method called something like `get_column_values` instead that there would be less confusion.

Regarding `__setitem__`, it is confusion to students that there is an `append` method that gets called to add rows but no similar method for columns. Instead, a largely mysterious assignment using bracket notation is needed.

    my_table['New Column'] = [some values...]

This makes sense to use because we commonly use dictionaries, but students have never used dicts before and so this is another thing they have to keep in their heads as they code.

Students basically have to memorize:

- Brackets on their own create lists.
- Use integers to retrieve items when using bracket notation for lists/arrays.
- Use integers + brackets to assign values for specific indices for lists/arrays.
- Use strings to retrieve values as an array when using bracket notation for tables.
- Use strings + brackets to create new columns for tables.

Even my strongest students struggle with the syntax of parentheses vs. brackets, and I think that their confusion is very justified given how inconsistent and seemingly arbitrary the usage of brackets is. (As icing on the cake, we can't document  `__getitem__` and `__setitem__` in the online docs because students don't know that these two methods are syntatic sugar for bracket notation.) We can do away with a lot of this by just replacing them with actual methods.

I would like to:

- Add a method called `get_column_values` that behaves like `__getitem__`.
- Add a method called `add_column` that behaves like `__setitem__`.
- Add a deprecation message when `__getitem__` and `__setitem__` are used, telling users to use the new methods instead.
- Remove the usage of `__getitem__` and `__setitem__` where we can (mainly in the text).
- Next semester, pretend they never existed.

Any thoughts on this? @papajohn @henryem @rjboczar @alvinwan

SamLau95:
Aliases `__getitem__` and `__setitem__` to these methods.

Resolves #80 

Also moves some methods around so that they are grouped by the groups in the docs.

```python
        >>> t
        letter | count | points
        a      | 9     | 1
        b      | 3     | 2
        c      | 3     | 2
        z      | 1     | 10

        >>> t.values("letter")
        array(['a', 'b', 'c', 'z'])
        >>> t.values("count")
        array([9, 3, 3, 1])
```

```python
        >>> table
        letter | count | points
        a      | 9     | 1
        b      | 3     | 2
        c      | 3     | 2
        z      | 1     | 10

        >>> table.append_column('new_col1', [10, 20, 30, 40])
        >>> table
        letter | count | points | new_col1
        a      | 9     | 1      | 10
        b      | 3     | 2      | 20
        c      | 3     | 2      | 30
        z      | 1     | 10     | 40
        >>> table.append_column('new_col2', 'hello')
        >>> table
        letter | count | points | new_col1 | new_col2
        a      | 9     | 1      | 10       | hello
        b      | 3     | 2      | 20       | hello
        c      | 3     | 2      | 30       | hello
        z      | 1     | 10     | 40       | hello

        >>> table.append_column(123, [1, 2, 3, 4])
        <ValueError raised>
        >>> table.append_column('bad_col', [1, 2])
        <ValueError raised>
```


deculler:
We have a subtle set of issues around the issue of columns labeled by an
integer, rather than a string.  I bumped into this fixing what appeared to
be a bug in _repr_html_.  It ought to wrap 'str' around the column label.
However, if we allow columns to be labeled with integers a lot of other
things break, since we use the fact that the argument is a string to
determine that a column is intended.  So, rather we need to be careful that
a table cannot be constructed with ints as column labels.  The place this
can pop up is pivot, where the column labels are derived from the values in
a column.  I ran into this working with some of the health data where all
the categorical stuff starts out as integer encodings.

David E. Culler

Issue #90, opened 2015-09-30
Documentation for `Table.take`


SamLau95:
Let's add more docs to `Table.take`! See `Table.split` for an example of a good docstring.

SamLau95:
Hey @taylorkmw, it's great to have you on board. Please take on this issue and submit a pull request! Ping me if you have any questions.

SamLau95:
Closed by #90. In the future, you can put a line like Resolves #90 in your PR / commit message and Github will autoclose the issue :)

PR #96, opened 2015-10-04
add docs for Table.take #90

taylorkmw:


SamLau95:
Good stuff! Thanks Taylor :)

Issue #58, opened 2015-09-09
Barh doesn't guarantee chart order


papajohn:
Charts for columns might come out in the wrong order, which is surprising. They should be rendered in column order.

SamLau95:
@ericz82 , wanna take this one?

ericz82:
Sure

ericz82:
Hmmm... I ran a few tests but the charts seem to be rendering in the order of the columns, unless I am supposed to looking at something else (maybe the rows)
[Untitled.html | uploaded via ZenHub](https://files.zenhub.io/5611b51065668d7d71c76ae5)

SamLau95:
Sorry for the confusion — we're referring to the fact that the rows don't get plotted in the right order. In your uploaded example, the bars should be a, b, c, then z (but they're z, c, b, then a).

SamLau95:
Closed by #98. In the future, you can put a line like Resolves #98  in your PR / commit message and Github will autoclose the issue :)

PR #98, opened 2015-10-05
fixed order of bars for barh

ericz82:
Turns out pyplot.barh plots bars starting from the bottom to the top

Fixes issue #58

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -888,10 +888,10 @@ class Table(collections.abc.MutableMapping):
                 ypos = index + margin + (1-2*margin)*labels.index(label)/len(labels)
             else:
                 ypos = index
-            axis.barh(ypos, self[label], width,  color=color, **options)
+            axis.barh(ypos, self[label][::-1], width,  color=color, **options)
         def annotate(axis, ticks):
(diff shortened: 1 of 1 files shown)


datascience/tables.py
```diff
@@ -888,10 +888,10 @@ def draw(axis, label, color):
                 ypos = index + margin + (1-2*margin)*labels.index(label)/len(labels)
             else:
                 ypos = index
-            axis.barh(ypos, self[label], width,  color=color, **options)
+            axis.barh(ypos, self[label][::-1], width,  color=color, **options)
         def annotate(axis, ticks):
             axis.set_yticks(index+0.5) # Center labels on bars
-            axis.set_yticklabels(ticks, stretch='ultra-condensed')
+            axis.set_yticklabels(ticks[::-1], stretch='ultra-condensed')
```
SamLau95: This is a bit unintuitive for someone reading the code for the first time. Can you add a comment here saying something like the default behavior is to display in reverse order which is why we reverse again here?

ericz82: Yeah, you want me to make another pull request?

SamLau95: No worries, just commit and push to master for small changes like that.

Issue #46, opened 2015-08-18
Table does not seem to support more than 255 columns


papajohn:
Here's a piece of code that triggers this issue:
```
lyricsTable = Table.read_table('http://eecs.berkeley.edu/~xinghao/ds10data/lyricsTable.csv')
lyricsTable
gives an error of
  File "<string>", line 12
SyntaxError: more than 255 arguments
```

If I used a smaller table (with 104 columns instead of 5004), the error goes away:
```
lyricsTable = Table.read_table('http://eecs.berkeley.edu/~xinghao/ds10data/lyricsTable_part.csv')
lyricsTable
```

papajohn:
I think the problem is my overuse of `*args` syntax. Probably we can fix it just by isolating the problem and passing arguments a different way.

deculler:
Be nice to handle it gracefully, but ...

5004 by what?  Thousands of lines?  That's a mighty table.  5004x1 ???

255 columns is perhaps a constraint if you think of it as a matrix.  But if
it is words and counts it should be two columns anyways.

papajohn:
These are counts of 5000 different words. Our table representation doesn't
do sparse matrices. It also doesn't handle 5000 columns very gracefully —
we'll have to think about that a little.

deculler:
But it is happy with columns of 5,000 entries, depending on how ordered
dict handles sparsity.

The natural view here is that each song is represented by a feature vector
of a bunch of words.  (5k is a BIG dictionary for lyrics).  You would
expect a 'word' column and one per song.  And the natural thing to do would
be to build the dictionary from songs, filtering on some frequency of use.
As more songs are considered, take the union of dictionaries.  Nice
alternative to join - a kind of open world, rather than closed world
perspective.  At any point, songs could be collapsed into genre, thereby
managing the number of columns.

In part we are seeing the distance between the mathematical sparse-matrix
view of the world and one that rests heavily on interacting with a useful
representation.

20,000 songs of 5,000 words - 100 million cells.  Very sparse indeed.  Not
immense, but not trivial either.

David E. Culler
Friesen Professor of Computer Science
Electrical Engineering and Computer Sciences
University of California, Berkeley

mijordan3:
Hmm, I'm used to seeing all kinds of data science problems with many tens
of thousands or even hundreds
of thousands of columns.  In some cases the matrices are sparse (along the
rows) but definitely not in all cases.

If the table implementation can be made to work for 5000 columns then
that's OK for this class, but if you
[...]

PR #101, opened 2015-10-06
Provide new minimal representation for row

mDibyo:
A row was earlier represented as a namedtuple. But this resulted in
the restriction that a table could not have more than 255 columns
(see http://stackoverflow.com/questions/18550270/any-way-to-bypass-namedtuple-255-arguments-limitation).

The new representation is a subclass of tuple which still
approximately provides access to column labels and has the same
representation. Additionally, column values can be accessed as
attributes and a mapping from column label to values can be obtained
using the asdict() method. Other behavior can be filled in as
required.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1316,17 +1316,17 @@ class Table(collections.abc.MutableMapping):
                 self._labels = labels
 
                 class Row(tuple):
-                    _table = self._table
+                    __table = self._table
 
                     def __getattr__(self, column_label):
-                        return self[self._table.column_index(column_label)]
+                        return self[self.__table.column_index(column_label)]
 
                     def __repr__(self):
                         return 'Row({})'.format(', '.join('{}={}'.format(
-                            self._table.column_labels[i], v.__repr__()) for i, v in enumerate(self)))
+                            self.__table.column_labels[i], v.__repr__()) for i, v in enumerate(self)))
 
                     def asdict(self):
-                        return collections.OrderedDict(zip(self._table.column_labels, self))
+                        return collections.OrderedDict(zip(self.__table.column_labels, self))
 
                 self._row = Row
 
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -581,7 +581,7 @@ class Table(collections.abc.MutableMapping):
         """
         rows, column_labels = [], column_labels or self.column_labels
         for row in self.rows:
-            [rows.append((getattr(row, key), k, v)) for k, v in row._asdict().items()
+            [rows.append((getattr(row, key), k, v)) for k, v in row.asdict().items()
              if k != key and k in column_labels]
         return Table.from_rows(rows, [key, 'column', 'value'])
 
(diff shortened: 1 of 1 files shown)


mDibyo:
This PR fixes #46.

mijordan3:
A bit more info on what we're thinking to do with the lyrics dataset:

Each row is a song.  The columns are the words, and the entry in a given
(row, column) is the number of times that
that word appears in the song.  (Perhaps normalized).  The final column is
the label of the genre.  We're doing two genres,
so it will be binary.

Part I:  To classify a song, we define a distance metric between rows,
presumably just Euclidean distance.
We then find the K nearest rows to the row representing the song we're
classifying, for K fixed a apriori (to, say 10).
Look up the class labels in the final column for those K nearest rows.
Take a vote over the binary labels
to label the query song.  Use a held-out test set to evaluate how well this
algorithm does.

Part II:  Use a third set of data, the cross-validation set, to try out
various values of K.  Find the winner in terms
of classification accuracy on the cross-validation set, and then fix that
value of K to classify the held-out test set.

Part III:  Find the single word that is most discriminative for this task.
I.e., from the original data, select a single
word column and redo Part I using that single column (together with the
label column).  Loop over the 5000 columns
and find the best predictor (on the cross-validation set).

So this project (and many similar projects in data science) require rows to
be first-class citizens.

Mike

deculler:
This looks great.  I'm so glad that you added asdict.  Conceptually, it is
natural to view rows as a 'record' as they used to say.

I did a one slide version of tables today for a talk at the Data Science
Environments workshop in the morning.  I've attached it here.  With this
PR, the little white lie went away.

David E. Culler

deculler:
If I've got it right, 5,000 is the words in the dictionary. You probably
want to think of each song as a column.  Not only can you name the columns
with the song names, it will be a couple orders of magnitude faster. And
you can control what you need to cope with by deciding how large a library
to include.  It is harder to leave out words.

Tables fundamentally are a ordered collection of named columns.  Columns
are numpy arrays.  Rows are not.  Not by a long ways.  And you have the
full power of numpy built in on columns.

Or for us old guys, think Fortran, not C.

David E. Culler

mijordan3:
We're going to be wanting to work with tens of thousands of songs.  Will
the representation support that
number of columns?

And will it be easy to loop over rows in the feature selection problem?

(I'm confused at this point, because I thought that we were working with a
paradigm in which data points are rows.
The songs are the data points in this problem.)

M

mDibyo:
The current representation will support unlimited columns and rows (within memory bounds. )

Looping over rows is possible (`for row in table.rows`) but is a slower process. Looping over rows is a faster process. To elaborate, the table currently is backed by an array of vectors, wherein each vector holds a column. (basically, as @deculler mentioned, Fortran and not C) A row is obtained by stringing together the corresponding value from each of these vectors. This is also part of the reason why we have a wrapper to represent the Row, while we don't have one for columns.

So, we would want our data points to be columns.

It might be possible to add a switch to Table which when 'on' makes our table an array of rows. This would not change the table interface, only change the way operations are handled internally. If done right, it would only require minor changes to the code. If we do this, we could support data points as rows. @papajohn, what do you think?

papajohn:
I think supporting multiple implementations will complicate table.py more
than I'd like, so I'd rather have row access be slow for now.

papajohn:
I think it should be no problem to loop over 10k rows many times in this implementation. merging.

deculler:
Agree on that.

PR #103, opened 2015-10-06
Clean versioning

stefanv:


SamLau95:
This is much better than what we had before! Will a version like `0.3.dev13` be considered more up-to-date than `0.3.12` transparently?

ryanlovett:
0.3.14 is already up on pypi and is greater than this commit's 0.3.dev13. (or is it?)

stefanv:
I guess someone forgot to update the version number.  I'll update it here.

stefanv:
Also see https://www.python.org/dev/peps/pep-0440/#developmental-releases

papajohn:
Nifty.

PR #104, opened 2015-10-06
Remove checked in docs build, ignore build output directory

stefanv:


SamLau95:
This will cause problems when Travis autobuilds our docs. See the [deploy_docs.sh](https://github.com/dsten/datascience/blob/master/deploy_docs.sh) (which is run by Travis) and [make deploy_docs](https://github.com/dsten/datascience/blob/master/Makefile) commands.

Travis commits the `docs/` folder, does a `git subtree split` on it, then pushes the html files to the gh-pages branch of this repo. With docs/_build in the gitignore, Travis won't be able to find those changes.

stefanv:
There's no need to commit the docs in order to have Travis upload them.  See e.g.

https://github.com/scikit-image/scikit-image/blob/master/.travis.yml#L60

papajohn:
I'm going to hold off on merging until @SamLau95 confirms that he has figured out how to deploy the docs in this new way.

stefanv:
Yup, makes sense, happy to help with that.

SamLau95:
Here goes nothing!

PR #105, opened 2015-10-06
Use __all__ to specify exposed API

stefanv:
Only expose implemented API so that introspection can be used in IPython to discover the module content.


SamLau95:
Didn't know about this. Thanks!

PR #106, opened 2015-10-06
Import datascience as ds

stefanv:
It is considered bad practice to do "from my_library import *".  This PR uses a more standard convention:

```
import datascience as ds
ds.Table(...)
```


SamLau95:
This is fine by me — I believe we do want to use that syntax in general to more easily talk about `Table` to students but doing `import as` for the tests is fine.

PR #109, opened 2015-10-06
Add conversion to/from Pandas DataFrame

stefanv:


datascience/tables.py
```diff
@@ -1298,6 +1297,16 @@ def points(self, column__lat, column__long, labels=None, colors=None, **kwargs)
...
 
+    @staticmethod
+    def from_df(df):
+        """Convert DataFrame into Table"""
+        labels = df.columns
+        return Table([df[label].values for label in labels], labels)
+
+    def to_df(self):
+        """Convert the table to a Pandas DataFrame"""
+        return pandas.DataFrame(self._columns)
```
SamLau95: Since we're exposing new methods on the `Table` class, could you add them to [`docs/tables.rst'](https://github.com/dsten/datascience/blob/master/docs/tables.rst) so we get them in the docs?

stefanv:
@SamLau95 Thanks, good feedback!  I made the changes.

PR #110, opened 2015-10-07
Add row slicing

stefanv:
This lets Tables behave similar to other Python containers, such that:

```
In [1]: import datascience as ds

In [2]: t = ds.Table((['a','b','c','d'], [1, 2, 3, 4]), ('names', 'counts'))

In [3]: print(t[:2])
names | counts
a     | 1
b     | 2

In [4]: print(t[-1])
names | counts
d     | 4

In [5]: print(t[1:-1])
names | counts
b     | 2
c     | 3

```

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -148,7 +148,11 @@ class Table(collections.abc.MutableMapping):
     #################
 
     def __getitem__(self, label):
-        return self.values(label)
+        if isinstance(label, slice) or hasattr(label, '__index__'):
+            return self._with_columns(column[label]
+                                      for column in self._columns.values())
+        else:
+            return self.values(label)
 
     def __setitem__(self, label, values):
         self.append_column(label, values)
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -749,3 +749,13 @@ def test_q_chaining(table):
     letter | count | points
     a      | 9     | 1
     """)
+
+def test_row_slicing(table):
+    t = table[:2]
+    assert_equal(t.column_labels, ('letter', 'count', 'points'))
+    assert_array_equal(t.columns[0], ['a', 'b'])
+    assert_array_equal(t.columns[1], [9, 3])
+    assert_array_equal(t.columns[2], [1, 2])
+
+    t_row = table[-1]
+    assert_equal(len(t_row.rows), 1)


SamLau95:
This syntax is a tad confusing because we've only had stuff like `t['label']` returning an array of the values of that column up to this point. @papajohn any thoughts?

stefanv:
Is there currently another way of easily grabbing a few rows? This is a
common enough operation that I presumed I must have missed it.

papajohn:
Nice suggestion, but tables are indexed by column name rather than number. You could change `t.columns()` to a special container that gives you a selected table upon slicing, rather than a list of arrays.

SamLau95:
@stefanv There's [Table.take](http://data8.org/datascience/_autosummary/datascience.tables.Table.take.html#datascience.tables.Table.take) which students have seen before.

stefanv:
Can take receive slices as input?

stefanv:
@papajohn This is an interesting design decision. Isn't it much more common
to slice out rows than to select columns?

papajohn:
In our course, the first table manipulation is to create new columns from existing columns. Named columns make the resulting expressions fairly easy to interpret.

E.g. http://data8.org/text/1_data.html#tables

Most of our examples don't pick out rows based on index, but instead based on their contents (e.g., using `where`) or by sampling.

papajohn:
@stefanv `take` can take a `range` rather than a slice. (We actually teach `np.arange` instead b/c Python 3 ranges take extra explanation.)

deculler:
take by index
where by value

Both produce a new table.

stefanv:
@papajohn Thanks for the link to the notes, I see now that the column syntax is often used for operations such as ``table['diff'] = table['2015'] - table['2012'``.

One way to do this would be to repurpose ``.rows``. We could, e.g., do ``t.rows[:15]`` and get a table back.  ``t.rows[0]`` (the current usage) will still be supported.

Indexing by range is probably fine for smaller queries, but becomes expensive for larger ones:

```
In [1]: import numpy as np

In [2]: x = np.random.random(100000)

In [3]: %timeit x[:50000]
The slowest run took 122.46 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 259 ns per loop

In [4]: z = np.arange(50000)

In [5]: %timeit x[z]
The slowest run took 7.63 times longer than the fastest. This could mean that an intermediate result is being cached 
10000 loops, best of 3: 97.4 µs per loop
```

Has a design decision been made on whether a `take` operation should always copy data, or whether tables can re-use underlying column storage?  Again, a trade-off between simplicity and memory usage (which can be limiting when working on large datasets).

stefanv:
See #120.

PR #111, opened 2015-10-07
WIP: Tutorial (do not merge)

stefanv:
This is a beginning for a tutorial.  Waiting on a decision for the slicing PR before continuing.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1,16 +1,66 @@
 Start Here: ``datascience`` Tutorial
 ====================================
 
-In progress.
+This is general introduction to the functionality in
+:py:mod:`datascience`.  For a reference guide, please see
+:ref:`tables-overview`.
 
 Introduction
 ------------
 
+First, load the :py:mod:`datascience` module:
+
+.. ipython:: python
+
+   import datascience as ds
+
+In the IPython notebook, type `ds.` followed by the TAB-key to see a list of members.
+The most important of these is the :py:class:`Table` class, which is
+the structure used to represent columns of data.
+
 Basic Table Usage
 -----------------
 
+A table is constructed as follows:
+
+.. ipython:: python
+
+   letters = ['a', 'b', 'c', 'z']
+   counts = [9, 3, 3, 1]
+   points = [1, 2, 2, 10]
+
+   t = ds.Table(columns=[letters, counts, points],
+                labels=['letter', 'count', 'points'])
+
+   print(t)
+
+Note how the first keyword, ``columns``, specifies the contents of the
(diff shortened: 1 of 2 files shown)


docs/tutorial.rst
```diff
@@ -1,16 +1,67 @@
...
+
+First, load the :py:mod:`datascience` module:
+
+.. ipython:: python
+
+   import datascience as ds
+
+In the IPython notebook, type `ds.` followed by the TAB-key to see a list of members.
+The most important of these is the :py:class:`Table` class, which is
+the structure used to represent columns of data.
```
SamLau95: We've always used `from datascience import *` with the students so they've never seen the `ds.Table` syntax before. We'll definitely want to avoid dropping an entirely new syntax on students since we're in the middle of the semester. I'm fine with doing something like this next semester if @papajohn is cool with it.

SamLau95:
Thanks a *ton* for starting this! Would love to know what @papajohn , @henryem , and @rjboczar would like to add as well.

docs/tutorial.rst
stefanv: This is considered poor practice in the Python community, so I'd strongly
recommend this syntax (same is being used by numpy, pandas, scipy,
matplotlib, etc.)

I understand about this semester potentially being problematic, though!

papajohn:
Just chiming in to say thanks! I'd recommend scanning through http://data8.org/text/1_data.html#tables to see how our students are currently being introduced to the `Table` class.

docs/tutorial.rst
SamLau95: Yup, I agree. Since this tutorial will be put online once its merged and students are actively referring to the docs right now, let's write the docs assuming we're going to use `from datascience import *` for now. We can comment out the docs that use `import datascience as ds` for now and then put them back in once this semester finishes up.

stefanv: Alright, will do!

stefanv:
Jessie, why close the WIP PR?

SamLau95:
Reopening since I don't think this was meant to be closed.

SamLau95:
@papajohn Hey, can we get your opinion on doing `ds.Table` versus just `Table`?

papajohn:
Let's write this for students. I'd prefer just `Table`.

stefanv:
If we use just Table, can we then also use the convention ``from datascience import Table`` etc. to show explicitly where these objects come from?  The ``from x import *`` convention is frowned upon for a reason: it clobbers the current namespace, and it prevents the use of checkers such as pyflakes from analysing code properly.

papajohn:
Sure, explicit imports in the tutorial are fine.

PR #115, opened 2015-10-08
Add property row for single-row tables

### LINKED ISSUE #93: Need a better way to extract the contents of one-row tables
After `t.sum()` or some other reduction to a single row, it's messy to access the contents of the result using `t.rows[0]...`.

mDibyo:
This resolves #93.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -202,6 +202,11 @@ class Table(collections.abc.MutableMapping):
         return self.Rows(self)
 
     @property
+    def row(self):
+        assert self.num_rows == 1, 'Table has more than one row.'
+        return self.rows[0]
+
+    @property
     def column_labels(self):
         """Return a tuple of column labels."""
         return tuple(self._columns.keys())
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -212,7 +212,8 @@ class Table(collections.abc.MutableMapping):
         Returns:
             The only ``Row`` in the table.
         """
-        assert self.num_rows == 1, 'Table has more than one row.'
+        if not self.num_rows == 1:
+            raise RuntimeError('Table has more than one row.')
         return self.rows[0]
 
     @property
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -203,6 +203,15 @@ class Table(collections.abc.MutableMapping):
 
     @property
     def row(self):
+        """Returns the only row in single-row tables. This can not not be used
+        when the table contains 0 or more than 1 rows.
+
+        Raises:
+            ``RuntimeError``: table has 0 or more than 1 row
+
+        Returns:
+            The only ``Row`` in the table.
+        """
         assert self.num_rows == 1, 'Table has more than one row.'
         return self.rows[0]
 
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -86,6 +86,23 @@ def test_basic_rows(t):
         "Row(letter='c', count=3, points=2)")
 
 
+def test_rows_slice(t):
+    rows = t.rows[1:]
+    assert_equal(len(rows), 3)
+    assert_equal(rows[0], "Row(letter='b', count=3, points=2)")
+
+
+def test_row():
+    t = Table([[1], [2]], ['one', 'two'])
+    assert_equal(t.row, 'Row(one=1, two=2)')
+
+
+def test_row_when_multiple_rows():
+    t = Table([[1, 3], [2, 4]], ['one', 'two'])
+    with pytest.raises(AssertionError):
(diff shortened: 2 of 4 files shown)


datascience/tables.py
```diff
@@ -202,6 +202,11 @@ def rows(self):
         return self.Rows(self)
 
     @property
+    def row(self):
+        assert self.num_rows == 1, 'Table has more than one row.'
+        return self.rows[0]
```
SamLau95: Let's write a docstring for this method! And then add it to `docs/tables.rst`.

mDibyo: Done!
`ValueError` does not seem the right choice, as its used to denote an error in input args. Switched to a `RuntimeError` instead.

stefanv:
This is a very specialized method to add, and only applicable to a small subset of tables.  I am working on another "proof of concept" PR that would add ``t.rows[0]`` as a syntax for this.  I'll make a PR later today, then we can discuss it over there, if that's of any interest.

mDibyo:
@stefanv, I am not sure I am following you. Are you working on a PR for an alternate way to get `table.rows[0]`?

tests/test_tables.py
```diff
@@ -86,6 +86,23 @@ def test_basic_rows(t):
...
+
+def test_row():
+    t = Table([[1], [2]], ['one', 'two'])
+    assert_equal(t.row, 'Row(one=1, two=2)')
+
+
+def test_row_when_multiple_rows():
+    t = Table([[1, 3], [2, 4]], ['one', 'two'])
+    with pytest.raises(RuntimeError):
+        _ = t.row
```
SamLau95: Thanks for writing these!

SamLau95:
Waiting on @papajohn and @stefanv to respond since I'm neutral about adding this method to the API.

stefanv:
Yes, my feeling (without knowing the background of the Row class) is that it is not really necessary.  I will soon file an exploratory PR that refactors ``rows`` to return a table, and then ``rows[0]`` with give you the single row required.  I don't think it's worth adding a new method *just* to return the first row.

Sorry, folks, if I am introducing noise into these conversations—I know I am not up to date with the discussions had so far, but I hope to catch up.

mDibyo:
@stefanv That's no problem! :)
Well `table.rows[0]` does indeed work right now. We are introducing `table.row` specifically for single-row table to make it easier to get the single row. Check out discussion at #93.
To get rows as a separate table, we already have `Table.take`. Do you think we should not have a separate method for getting just individual data points (not in table format)?

tests/test_tables.py
mDibyo: Ohhh, don't thank me. I should always be writing tests for any method I write...

deculler:
John had pretty strong feelings about the row design, so I'll let him weigh
in here.  Seems like you are working with the Tables_as_matrix concept
rather than ordered collection of named columns.  A row is really like a
record, or rather Dict.  But of course it can be a sequence too, since the
columns are ordered.  This also preserves the connection with relations -
tables in SQL.  So I would say the design is not a random accident.  It was
pretty carefully developed as a concept.  The teaching goal helps keep that
in focus.  Yes there are lots of implementations, and the row
implementation has gone through some changes.

stefanv:
The way I think about it, a Table is a container of columns.  If you then extract entries from the table, you should again get a multi-column object (which, essentially, is a Table again).  If you think of a Table as a container of rows, the abstraction described above fits more easily.  But I think trying to accommodate both can be quite confusing.  So, what I'm implementing is a fairly minor change that simply removes the notion of a Row, but still gives you a Table object when selecting a row—and that object still has access to columns, as expected.

stefanv:
After some more experimenting, the above seems to be doable using array scalars, but it does break down on the notion of ``table.rows[0]`` giving a single object and ``table.rows[:2]`` returning a container (as you would get in, e.g., numpy and other N-D containers).  So perhaps then the smallest change that would make sense is to simply alter ``table.rows`` to accept a slice and then return a Table instead of multiple rows.

stefanv:
That approach turned out to be a *lot* easier: https://github.com/dsten/datascience/pull/119

Which makes me think it's perhaps closer to the right pattern.

stefanv:
See also #120 for a different idea (adding slicing to `take`)

deculler:
I definitely like the idea of adding slicing to take.  It is frustrating to
have to write the enumerator there.

David E. Culler

SamLau95:
Pinging @papajohn ; I agree that this method is very specialized and it's always much harder to remove a method than to add one. Do you see this being useful enough to add it to the API?

papajohn:
Let's drop the single-row tables special support.

PR #117, opened 2015-10-09
Add classmethods for creating empty tables and tables from a dict of column labels to column values

### LINKED ISSUE #88: Need a better way to define an empty table
`Table([[]], ['sample'])` is too clunky. Maybe `Table.empty(['sample'])`. Thoughts?

mDibyo:
This resolves #88.

The documentation for Table.__init__ has been simplified quite a bit by extracting the use case of initializing a table from a dict of column labels to column values into Table.from_columns_dict.

Also added a Table.empty method to create an empty table.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -30,7 +30,7 @@ class Table(collections.abc.MutableMapping):
     ##########
 
     def __init__(self, columns=None, labels=None,
-            formatter=_formats.default_formatter):
+                 formatter=_formats.default_formatter):
         """Create a table from a list of column values or dictionary of
         sequences.
 
@@ -45,23 +45,16 @@ class Table(collections.abc.MutableMapping):
         c      | 3     | 2
         z      | 1     | 10
 
-        For other ways to initialize a table, see :func:`Table.from_rows`,
-        :func:`Table.from_records`, and :func:`Table.read_table`.
+        For other ways to initialize a table, see ;func;`Table.empty`,
+        :func:`Table.from_rows`, :func:`Table.from_records`,
+        :func:`Table.read_table`, and :func:`Table.from_columns_dict`.
 
         Kwargs:
-            columns (None, list, or dict): If ``None``, an empty table is
-                created.
-
-                If a list, each element in ``columns`` is another list
+            columns (list): A list in which each element is another list
                 containing the values for a column in the order the columns
                 appear in ``labels``.
 
-                If a dict, each key is a label of a column; each values is the
-                column's values as a list.
-
-            labels (list): A list of column labels. Must be specified if
-                ``columns`` is a list, and must be ``None`` if ``columns`` is a
-                dict.
+            labels (list): A list of column labels.
 
             formatter (Formatter): An instance of :class:`Formatter` that
                 formats the columns' values.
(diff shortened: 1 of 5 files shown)


datascience/tables.py
```diff
@@ -45,23 +45,16 @@ def __init__(self, columns=None, labels=None,
         c      | 3     | 2
         z      | 1     | 10
 
-        For other ways to initialize a table, see :func:`Table.from_rows`,
-        :func:`Table.from_records`, and :func:`Table.read_table`.
+        For other ways to initialize a table, see ;func;`Table.empty`,
```
SamLau95: This should be `:func:` instead of `;func;`, right?

datascience/tables.py
```diff
@@ -82,13 +75,8 @@ def __init__(self, columns=None, labels=None,
...
         # Ensure inputs are properly formed
-        if not columns:
-            assert not labels, 'labels but no columns'
-            columns, labels = [], []
-        if isinstance(columns, collections.abc.Mapping):
-            assert labels is None, 'labels must be None if columns has labels'
-            columns, labels = columns.values(), columns.keys()
-        assert labels is not None, 'Labels are required'
+        columns = columns if columns is not None else []
+        labels = labels if labels is not None else []
```
SamLau95: Nice!

datascience/tables.py
mDibyo: Oops! Now corrected... Nice catch!

datascience/tables.py
```diff
@@ -98,6 +86,14 @@ def __init__(self, columns=None, labels=None,
             self[label] = column
 
     @classmethod
+    def empty(cls, column_labels=None):
+        """Create an empty table."""
```
SamLau95: Let's document this method a little better; just describing the args and return value would be very helpful for someone wondering how to use it.

datascience/tables.py
```diff
@@ -111,6 +107,11 @@ def from_records(cls, records):
         return cls([[rec[label] for rec in records] for label in labels], labels)
 
     @classmethod
+    def from_columns_dict(cls, columns):
+        """Create a table from a mapping of column labels to column values."""
```
SamLau95: This one too!

SamLau95:
See my notes about writing docs. We're going to try our best to write docs as we go since it's easier to write them now than to write docs for a bunch of methods later.

papajohn:
How about adding this doctest example

```
>>> t = Table.empty(['x', 'y'])
>>> print(t.append((2, 3)))
x    | y
2    | 3
```

datascience/tables.py
mDibyo: Done.

datascience/tables.py
mDibyo: Done.

mDibyo:
While messing around with the doctests, I also realized that the doctests for many of the functions are not running. Will create a separate PR for that.

SamLau95:
Great! Thanks @mDibyo .

mDibyo:
@SamLau95 No problem! Could you link me to your notes for writing docs?

mDibyo:
Added PR for doctest fixes: #125

PR #119, opened 2015-10-10
POC: Slice rows into Table

stefanv:
This changes the semantics of array *slicing* to return a Table, being consistent with the notion of a Table consisting of multiple Rows.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1355,7 +1355,8 @@ class Table(collections.abc.MutableMapping):
 
         def __getitem__(self, i):
             if isinstance(i, slice):
-                return [self[j] for j in range(*i.indices(len(self)))]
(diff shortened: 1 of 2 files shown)


stefanv:
Here's the analogous NumPy slicing operation:

```
In [1]: z = np.array([[1, 2, 3], [4, 5, 6]])

In [2]: z[0]
Out[2]: array([1, 2, 3])

In [3]: z[:2]
Out[3]: 
array([[1, 2, 3],
       [4, 5, 6]])

In [4]: z[0].shape
Out[4]: (3,)

In [5]: z[:2].shape
Out[5]: (2, 3)
```

stefanv:
If you'd rather leave the row semantics alone, another option would be to change ``.take`` to be a special object (still supporting it's current invocation), but also adding

```
t.take[:10]
```

stefanv:
Closing this in favor of #120.

PR #120, opened 2015-10-10
POC: Allow take to handle slicing

stefanv:
Proof of concept: change `take` to handle slicing.

See also #119 and #115

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -37,7 +37,7 @@ class _Taker:
         Returns:
             A ``Table`` containing only the selected rows.
 
-        >>> print(t)
+        >>> t
         letter grade | gpa
         A+           | 4
         A            | 4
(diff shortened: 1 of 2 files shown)


SamLau95:
This is a cool extension of the API! I'm concerned about documenting this for students to use. Right now `Table.take` is documented and the removal of that (since `_Taker` is private) might not allow `Table.take` to be documented as well.

As an aside: This functionality probably wouldn't be taught to students (who already have a lot of trouble differentiating between when to use brackets and parens).

stefanv:
We could teach the next cohort to *only* use square brackets for indexing
(colums, take, etc.)

Good point about the documentation, although I think I can fix that fairly
easily. I'll give it a shot and update this PR!

mDibyo:
@stefanv, could you go into more detail about why you implemented a separate __getitem__ method instead of just adding handling of slices to take.__call__ (and sticking to the original method definition)?

Sorry if this is a stupid question. :D

stefanv:
Sure, it's because you cannot do ``t.take(:14)``.

stefanv:
Updated to alias ``take()`` to ``take[]``

stefanv:
The docstring also seems to be fine, at least from inside IPython.

SamLau95:
Does `Table.take` still show up on the docs after running `make docs`?

stefanv:
Fix the Sphinx doc.

stefanv:
Now depends on PR #127

SamLau95:
Sorry for the delay on merging this! So the verdict on this is that:

- The previous behavior of `take` is preserved, and we will teaching students that syntax (eg. `table.take(range(5))`).
- The new behavior allows for more succinct taking of rows from a table.

To that end, could we write examples that call `take` using the old syntax first in the docs? Then, we can "translate them" to the indexing/slicing syntax with a note that the two are equivalent. The reason for this is that students will be looking at these docs to learn how to use the method and I'd err on the side of being clear.

stefanv:
I've updated the documentation—let me know if that is what you had in mind.

SamLau95:
This looks great! Thanks for the hard work on this.

PR #122, opened 2015-10-10
Re-expose as_text and as_html. Fix version import for docs build.

stefanv:
Previously, I turned `as_text` and `as_html` into private functions, but I see they are documented and therefore public.  This PR restores those, and fixes an import in the docs build to ensure that the docs are built from the currently checked out source.


SamLau95:
Great! Thanks.

PR #124, opened 2015-10-10
Expose polygons for a Region

papajohn:


stefanv:
The only undocumented method added is ``properties``—would that be enough to trigger the failure?

datascience/maps.py
```diff
@@ -454,13 +457,50 @@ class Region(_MapFeature):
...
+        elif self.type == 'MultiPolygon':
+            polygons = self._geojson['geometry']['coordinates']
+        return [   [   [_lat_lons_from_geojson(s) for
+                        s in ring  ]              for
+                    ring in polygon]              for
+                polygon in polygons]
+
+    @property
+    def area(self):
+        """The area of the region."""
```
SamLau95: Was this meant to be blank?

papajohn: Removed blank method.

PR #125, opened 2015-10-11
Fix table doctests and add test_doctests to all test_*.py

mDibyo:
Title is self-explanatory.

These are mostly documentation changes, but I did have to change Table.append_column a bit to match expected behavior. Please check.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -874,7 +874,7 @@ class Table(collections.abc.MutableMapping):
         Returns:
             A tuple containing two instances of ``Table``.
 
-                >>> job = ['a', 'b', 'c', 'd']
+        >>> job = ['a', 'b', 'c', 'd']
         >>> wage = [10, 20, 15, 8]
         >>> foo_table = Table([job, wage], ['job', 'wage'])
         >>> print(foo_table)
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -475,6 +475,10 @@ class Table(collections.abc.MutableMapping):
         Returns:
             An instance of ``Table`` containing only selected columns.
 
+        >>> burgers = ['cheeseburger', 'hamburger', 'veggie burger']
+        >>> prices = [6, 5, 5]
+        >>> calories = [743, 651, 582]
+        >>> t = Table([burgers, prices, calories], ['burgers', 'prices', 'calories'])
         >>> print(t)
         burgers       | prices | calories
         cheeseburger  | 6      | 743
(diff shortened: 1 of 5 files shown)


datascience/tables.py
```diff
@@ -243,26 +243,29 @@ def values(self, label):
 
         ``__getitem__`` is aliased to this, so bracket notation can be used:
 
+        >>> letter = ['a', 'b', 'c', 'z']
+        >>> count = [9, 3, 3, 1]
+        >>> points = [1, 2, 2, 10]
+        >>> t = Table([letter, count, points], ['letter', 'count', 'points'])
         >>> t.values('letter') == t['letter']
-        True
+        array([ True,  True,  True,  True], dtype=bool)
```
SamLau95: Right, I forgot about this behavior. Do you think this is confusing if all we're trying to show is that `t.values` is the same as `t['letter']`?

datascience/tables.py
```diff
@@ -243,26 +243,29 @@ def values(self, label):
...
+        >>> print(t)
         letter | count | points
         a      | 9     | 1
         b      | 3     | 2
         c      | 3     | 2
         z      | 1     | 10
-
-        >>> t.values("letter")
-        array(['a', 'b', 'c', 'z'])
+        >>> t.values("letter") # doctest: +NORMALIZE_WHITESPACE
```
SamLau95: Interesting! What does `+NORMALIZE_WHITESPACE` do?

datascience/tables.py
```diff
@@ -243,26 +243,29 @@ def values(self, label):
...
+        array(['a', 'b', 'c', 'z'], dtype='<U1')
         >>> t.values("count")
         array([9, 3, 3, 1])
+
+        Args:
+            label (str): The name of the column
+
+        Returns:
+            An instance of ``numpy.array``.
```
SamLau95: Thanks for catching this.

datascience/tables.py
```diff
@@ -347,38 +350,44 @@ def append_column(self, label, values):
...
+        Traceback (most recent call last):
+            ...
+        ValueError: Column length mismatch. New column does not have the same number of rows as table.
         """
         # TODO(sam): Allow append_column to take in a another table, copying
         # over formatter as needed.
         if not isinstance(label, str):
             raise ValueError('The column label must be a string, but a '
-                '{} was given'.format(str.__class__.__name__))
+                '{} was given'.format(label.__class__.__name__))
```
SamLau95: Oops! Good catch.

mDibyo: Thanks!

datascience/tables.py
mDibyo: It treats all sequences of whitespace (blanks and newlines) equal. Any sequence of whitespace within the expected output will match any sequence of whitespace within the actual output. :) https://docs.python.org/3.4/library/doctest.html#doctest.NORMALIZE_WHITESPACE

datascience/tables.py
mDibyo: Yeah. I think you have got a point. I could replace this with 
```python
>>> t.values('letter') == t['letter'] # doctest: +SKIP
<True>
```

mDibyo:
Given that the doctests did expose a couple of discrepancies between documentation and code, I think we should be adding them as one of the Travis CI checks.
Of course, this would result in additional code writing and documentation in order to have perfectly-executable code in doctests.

stefanv:
@mDibyo I agree that it's best to have these tests run by Travis-CI.  Perhaps it is as simple as adding `py.test --doctest-modules`?

mDibyo:
@stefanv Haha, I didn't know you could just do that! Will create a PR for this. :D
What I meant was, everyone would have to spend more time making sure their doctests passed, and writing more code to make them pass before pushing.

datascience/tables.py
SamLau95: Actually, let's not have this in the doctests but instead just write it in the description for the method. Something like: `t.values('letter')` is equivalent to `t['letter']`.

SamLau95:
Great stuff! Will merge after my nit is addressed :)

mDibyo:
Done!

mDibyo:
I will add a Travis CI doctest check after fixing doctests in the other files, in a future PR.

mDibyo:
Scratch that! Doctests in all other modules already pass.

`py.test --doctest-modules` was giving me errors, so I added test_doctests to the respective test_*.py files instead.

Issue #118, opened 2015-10-09
Add with_relabeling and with_column methods


SamLau95:
See #114 for rationale.

`Table.with_relabeling` should behave like `Table.relabel` but returns a new table instead of mutating the original.

`Table.with_column` should do the same for `Table.append_column`.

SamLau95:
@ericz82 Hey, wanna take this on?

ericz82:
Should I make both of these class methods or instance methods?

PR #126, opened 2015-10-12
added methods Table.with_relabeling and Table.with_column

ericz82:
#118

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -396,26 +396,152 @@ class Table(collections.abc.MutableMapping):
         return self
 
     def with_column(self, label, values):
-        """Returns a new table with new column included."""
-        new_table = Table(self.columns, self.column_labels)
+        """Returns a separate new table with new column included.
+
+        Args:
+            ``label`` (str): The label of the new column.
+                Must be a string.
+
+            ``values`` (single value or list/array): If a single value, every
+                value in the new column is ``values``.
+
+                If a list or array, the new column contains the values in
+                ``values``. ``values`` must be the same length as the table.
+
+        Returns:
+            A new separate table containing new column specified by ``label``
+            and ``values`` in addition to previous columns.
+
+        Raises:
+            ``ValueError``: If
+                - ``label`` is not a string.
+                - ``values`` is a list/array and does not have the same length
+                  as the number of rows in the table.
+
+        >>> table
+        letter | count | points
+        a      | 9     | 1
+        b      | 3     | 2
+        c      | 3     | 2
+        z      | 1     | 10
+
+        >>> table2 = table.append_column('new_col1', [10, 20, 30, 40])
+        >>> table3 = table2.append_column('new_col2', 'hello')
+        >>> table2
+        letter | count | points | new_col1
+        a      | 9     | 1      | 10
+        b      | 3     | 2      | 20
+        c      | 3     | 2      | 30
+        z      | 1     | 10     | 40
+        >>> table3
+        letter | count | points | new_col1 | new_col2
+        a      | 9     | 1      | 10       | hello
+        b      | 3     | 2      | 20       | hello
+        c      | 3     | 2      | 30       | hello
+        z      | 1     | 10     | 40       | hello
+        >>> table
+        letter | count | points
+        a      | 9     | 1
+        b      | 3     | 2
+        c      | 3     | 2
+        z      | 1     | 10
+
+        >>> table.append_column(123, [1, 2, 3, 4])
+        <ValueError raised>
+        >>> table.append_column('bad_col', [1, 2])
+        <ValueError raised>
+        """
+        new_table = self.copy()
         new_table.append_column(label, values)
         return new_table
 
 
     def relabel(self, column_label, new_label):
-        """Change the label of a column."""
-        assert column_label in self._columns
-        rewrite = lambda s: new_label if s == column_label else s
+        """Change the labels of columns specified by ``column_label`` to
+        labels in ``new_label``.
+
+        Args:
+            ``column_label`` (single str or list/array of str): The label(s) of
+                columns to be changed. Must be str.
+
+            ``new_label`` (single str or list/array of): The new label(s) of
+                columns to be changed. Must be str.
+
+                Number of elements must match number of elements in
+                ``column_label``.
+
+        Returns:
+            None
+
+        >>> table = Table([(1, 2, 3), (12345, 123, 5123)], ['points', 'id'])
+        >>> table.relabel('id', 'yolo')
+        >>> table
+        points | yolo
+        1      | 12345
+        2      | 123
+        3      | 5123
+        >>> table.relabel(['points', 'yolo'], ['red', 'blue'])
+        >>> table
+        red    | blue
+        1      | 12345
+        2      | 123
+        3      | 5123
+        """
+        if isinstance(column_label, str) and isinstance(new_label, str):
+            column_label, new_label = [column_label], [new_label]
+        d = dict(zip(column_label, new_label)) # dictioanry to map old labels to new ones
+        for col in d.keys():
+            assert col in self._columns
+        rewrite = lambda s: d[s] if s in d.keys() else s
         columns = [(rewrite(s), c) for s, c in self._columns.items()]
         self._columns = collections.OrderedDict(columns)
-        if column_label in self._formats:
-            formatter = self._formats.pop(column_label)
-            self._formats[new_label] = formatter
+        for label in self._formats:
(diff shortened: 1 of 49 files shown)


SamLau95:
Cool! Couple things:

- Let's add documentation and tests for these methods
- Let's make these methods also be able to take in a sequence as well as an individual value. (Eg. `with_relabeling(['A', 'B', 'C'], ['a', 'b', 'c'])` should relabel all the columns specified.)

ericz82:
How do I add tests?

mDibyo:
you can add tests at [tests/test_tables.py](https://github.com/dsten/datascience/blob/master/tests/test_tables.py)

datascience/tables.py
```diff
@@ -395,6 +395,13 @@ def append_column(self, label, values):
         self._columns[label] = values
         return self
 
+    def with_column(self, label, values):
+        """Returns a new table with new column included."""
```
stefanv: This docstring is rather vague.

stefanv: Also, what does it add above `append_column`?  If this is just to make a copy, we should rather implement ``table.copy()``.

datascience/tables.py
```diff
@@ -406,6 +413,12 @@ def relabel(self, column_label, new_label):
             self._formats[new_label] = formatter
         return self
 
+    def with_relabeling(self, column_label, new_label):
+        """Returns a new table with column_label changed to new_label."""
```
stefanv: This docstring should be expanded along the lines of the others.  It should describe that the parameters expected can be a single string, or a tuple of strings.  Perhaps it can also take a dict?  Alternatives to the name ``with_relabeling`` could be ``table.relabel``, ``table.rename`` or ``table.rename_columns``.

docs/_autosummary/datascience.tables.Table.as_html.rst
```diff
@@ -0,0 +1,6 @@
+datascience.tables.Table.as_html
```
stefanv: These files should not be included in the PR.

ericz82:
Should I change Table.relabel to also take in a sequence of arguments?

stefanv:
@ericz82 I think it already does (according to the example you gave).  It should simply be documented that way.

ericz82:
@stefanv Hmm, I'm not sure what example you are referring to. I wrote a test for Table.relabel taking in a list for ``column_label`` and a list for ``new_label`` which does not seem to be passing. I believe Table.relabel should be changed accordingly.

stefanv:
Ah!  But there you go.  You should just rework `relabel` to have the functionality you want, instead of adding a whole new method.

ericz82:
Yeah, that's what I decided on doing. Everything should be good now.

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
+        a      | 9     | 1      | 10       | hello
+        b      | 3     | 2      | 20       | hello
+        c      | 3     | 2      | 30       | hello
+        z      | 1     | 10     | 40       | hello
+        >>> table
+        letter | count | points
+        a      | 9     | 1
+        b      | 3     | 2
+        c      | 3     | 2
+        z      | 1     | 10
```
SamLau95: Good that you showed that `table` is unchanged! But you wrote `table.append_column('new_col1', [10, 20, 30, 40])` instead of `table.with_column('new_col1', [10, 20, 30, 40])`. :)

docs/tables.rst
```diff
@@ -74,7 +74,9 @@ Mutation (modifies table in place)
     Table.move_to_end
     Table.append
     Table.append_column
+    Table.with_column
     Table.relabel
+    Table.with_relabeling
```
SamLau95: These belong in the Transformation section because they return a new table instead of mutating the original!

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
         self._columns[label] = values
         return self
 
+    def with_column(self, label, values):
+        """Returns a separate new table with new column included.
```
SamLau95: This can just say `Returns a new table with new column included.`

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
-        assert column_label in self._columns
-        rewrite = lambda s: new_label if s == column_label else s
+        """Change the labels of columns specified by ``column_label`` to
+        labels in ``new_label``.
+
+        Args:
+            ``column_label`` (single str or list/array of str): The label(s) of
+                columns to be changed. Must be str.
+
+            ``new_label`` (single str or list/array of): The new label(s) of
```
SamLau95: nit: `list/array of str`

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
+                columns to be changed. Must be str.
+
+                Number of elements must match number of elements in
+                ``column_label``.
```
SamLau95: If the lengths don't match up, let's throw a ValueError. See `Table.split` for an example.

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
+
+        Returns:
+            None
```
SamLau95: This actually returns the original table, so let's write that here.

docs/tables.rst
ericz82: @SamLau95 Should I also change the locations of these methods in tables.py?

stefanv:
So, API question: is the new method added here then equivalent to:

```
t1= t0.copy()
t1.relabel(['a','b','c'], ['A','B','C'])
```

?  If so, is it worth expanding the API? (I have no idea how common such an operation is.)

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
+        >>> table
+        red    | blue
+        1      | 12345
+        2      | 123
+        3      | 5123
+        """
+        if isinstance(column_label, str) and isinstance(new_label, str):
+            column_label, new_label = [column_label], [new_label]
+        d = dict(zip(column_label, new_label)) # dictioanry to map old labels to new ones
+        for col in d.keys():
```
SamLau95: Let's call this dictionary `old_to_new` instead of `d` since it'll help someone who's reading this for the first time. Also, dictionaries iterate over their keys by default so you can more succinctly write `for col in old_to_new:`.

nit: You misspelled `dictioanry`!

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
+            assert col in self._columns
```
SamLau95: A ValueError is be better to raise here!

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
+        rewrite = lambda s: d[s] if s in d.keys() else s
```
SamLau95: nit: As with above, you can omit the `.keys()`

datascience/tables.py
```diff
@@ -395,17 +395,156 @@ def append_column(self, label, values):
...
-            self._formats[new_label] = formatter
+        for label in self._formats:
+            if label in column_label:
+                formatter = self._formats.pop(label)
+                self._formats[d[label]] = formatter
         return self
 
+    def with_relabeling(self, column_label, new_label):
+        """Returns a new table with column_label changed to new_label.
+        ``column_label`` and ``new_label`` may be single values or lists
```
SamLau95: Let's insert a newline above the line starting with column_label and new_label so that the summary on the main table doc page is short.

docs/tables.rst
SamLau95: Good catch! Please do that.

SamLau95:
@stefanv I think it is worth expanding the API in this case because method chaining is much easier with these two methods. In addition, I believe we're going to deprecate the mutation methods (or at least not teach them) in favor of these new ones because of the issues we face with students running cells multiple times. See https://github.com/dsten/datascience/issues/114

stefanv:
@SamLau95 Thanks, that makes sense.  I think there are a couple of sneaky places in which data mutation can still accidentally happen.  If this is a decided policy (never to mutate), then we can start hunting for those.

SamLau95:
@stefanv I'm not sure if it's a decided policy but it definitely helps a lot with reasoning about tables, especially for students that are just starting to learn programming. Originally we were fine with mutation but I think the de facto policy now is just to replace mutation with a pure function alternative as much as we can.

SamLau95:
Thanks for the hard work on this @ericz82 !

PR #128, opened 2015-10-13
Use __str__ as default __repr__

stefanv:
At the moment, a person needs to type `print(t)` every time you want to see what's going on inside a table, which gets rather tedious.

This patch proposes to set ``__repr__`` to `__str__`` so that you get a more useful representation by default.


SamLau95:
Cool! I like this.

mDibyo:
this looks good. Will make the change to #125.

PR #129, opened 2015-10-13
Add conversion to/from NumPy structured arrays

stefanv:
Note that this PR also removes ``t.matrix``, which returned a somewhat broken matrix object (matrices are not recommended for use, and in this case the column dtype was simply being taken from the first column, turning numbers into strings, etc.)


SamLau95:
Looks good! What's the most common use-case for this?

stefanv:
I feel it's important to have bridges to well established tools that folks are familiar with already.  NumPy is certainly a de-facto standard!  Once you have a numpy array, you can save the file to disk for example, and load it back up.  A whole lot of standard tools become available to you.

SamLau95:
Cool!

PR #132, opened 2015-10-18
Correct Table.hist label when counts arg is provided

### LINKED ISSUE #112: Horizontal axis label should change when counts argument is provided to hist
With `counts=col`, the horizontal axis contains values from `col` not from the column plotted (which contains counts), but currently the label is from the column plotted.

mDibyo:
This fixes #112.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1580,11 +1580,13 @@ class Table(collections.abc.MutableMapping):
                 bins = np.unique(self[bins])
             vargs['bins'] = bins
 
-        counted_values = None
+        counted_values = counted_label = None
         if counts is not None:
             counted_values = self._get_column(counts)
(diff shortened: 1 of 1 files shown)


SamLau95:
Looks good! Do you think the lowered coverage is worth improving?

mDibyo:
I added a doctest for it. Since the changes are visual (function returns None), the only way we could test it is through visual testing.

Either that, or we could modify the function to return all the plot objects. We should do that for all the plotting functions, but this PR doesn't seem to be the right place for it. I will create an issue for that.

In that case, the only way to improve coverage would be to add tests for some other parts, but don't want to add those to this pull request.

SamLau95:
Right, that's very reasonable.

PR #153, opened 2015-11-28
fix scatter axes labels #146

taylorkmw:

Issue #146, opened 2015-11-02
Scatter should label axes correctly


papajohn:

SamLau95:
Here's an example of the current incorrect behavior:

```python
>>> a = Table([[10, 20, 30, 40, 50, 60, 70, 80,],
               [2, 13, 8, 5, 3, 2, 0, 1,],
               [3, 10, 13, 12, 2, 0, 0, 0,]],
              ['Age', 'Male Counts', 'Female Counts'])
>>> print(a)
Age  | Male Counts | Female Counts
10   | 2           | 3
20   | 13          | 10
30   | 8           | 13
40   | 5           | 12
50   | 3           | 2
60   | 2           | 0
70   | 0           | 0
80   | 1           | 0
>>> a.scatter('Age')
```
![screenshot 2015-11-24 13 35 26](https://cloud.githubusercontent.com/assets/2468904/11381329/441f1a94-92b0-11e5-8940-2432bcd1a296.png)

As you can see, the x-axis for these scatterplots should be 'Age' and the other column should be the label for the y-axis.

@taylorkmw Do you want to take this on?

PR #154, opened 2015-11-29
Modify _visualize to accept x and y labels as args #146

taylorkmw:
#153 already fixed the issue where the y-axis label was placed in the x-axis label's place for methods utilizing `visualize`. But this pull request revises `_visualize`'s method parameters to improve clarity. All the methods that currently use `_visualize` (`plot`, `scatter`, `bar`, and `barh`) use both x and y axes labels anyway, so  `_visualize` now accepts `x_label` and `y_labels` as params instead of simply `labels`.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1316,8 +1316,7 @@ class Table(collections.abc.MutableMapping):
         options = self.default_options.copy()
         options.update(vargs)
 
-        xticks, labels = self._split_by_column(column_for_xticks)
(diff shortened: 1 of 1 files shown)


SamLau95:
Thanks for doing this! Can you update the description of the PR to reference your previous PR? Ideally I should be able to tell not only what the PR is doing but also why it's here (eg. context and, if applicable, next steps). After that, this looks good to go :)

taylorkmw:
Done!

Issue #130, opened 2015-10-16
Apply should take multiple column labels


deculler:

SamLau95:
This is an example of the behavior we'd like:

```python
>>> a = Table([[10, 20, 30, 40, 50, 60, 70, 80,],
               [2, 13, 8, 5, 3, 2, 0, 1,],
               [3, 10, 13, 12, 2, 0, 0, 0,]],
              ['Age', 'Male Counts', 'Female Counts'])
>>> print(a)
Age  | Male Counts | Female Counts
10   | 2           | 3
20   | 13          | 10
30   | 8           | 13
40   | 5           | 12
50   | 3           | 2
60   | 2           | 0
70   | 0           | 0
80   | 1           | 0

>>> a.apply(lambda x: x * 2, 'Age') # This behavior already exists
array([ 20,  40,  60,  80, 100, 120, 140, 160])

# We want this behavior
>>> a.apply(lambda x, y: x + y, ['Male Counts', 'Female Counts'])
array([ 5, 23, 21, 17,  5,  2,  0,  1])
```

@ericz82 Do you wanna take this one?

ericz82:
Sure I'll work on this

PR #155, opened 2015-11-29
Changed Table.apply to take in multiple columns

ericz82:
Before, `Table.apply` could only apply a function with one argument to a single column in the table.

With this PR, we add the ability for `Table.apply` to take in an array of column names. See #130

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -447,6 +447,21 @@ class Table(collections.abc.MutableMapping):
         """Returns an array where ``fn`` is applied to each set of elements
         by row from the specified columns in ``column_label``.
 
+        Args:
+            ``fn`` (function): The function to be applied to elements specified
+                by ``column_label``.
+            ``column_label`` (single string or list of strings): Names of
+                columns to be passed into function ``fn``. Length must match
+                number of elements ``fn`` takes.
+
+        Raises:
+            ``ValueError``: column name in ``column_label`` is not an existing
+                column in the table.
+
+        Returns:
+            A numpy array consisting of results of applying ``fn`` to elements
+            specified by ``column_label`` in each row.
+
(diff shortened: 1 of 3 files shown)


datascience/tables.py
```diff
@@ -444,9 +444,26 @@ def column_index(self, column_label):
...
+        >>> points = [1, 2, 2, 10]
+        >>> t = Table([letter, count, points], ['letter', 'count', 'points'])
+        >>> t
+        letter | count | points
+        a      | 9     | 1
+        b      | 3     | 2
+        c      | 3     | 2
+        z      | 1     | 10
+        >>> t.apply(lambda x, y: x * y, ['count', 'points'])
+        array([ 9,  6,  6, 10])
```
SamLau95: Can we use the same documentation format as `Table.split`? Might as well leave the documentation better than how we found it when we can :)

It's important to mention that the `column_label` argument can take either a string or a list of column names.

In addition, we can throw an intelligent error when a column name passed in is not in the table — right now the student will only see a cryptic KeyError. See `Table.split` for an example.

tests/test_tables.py
```diff
@@ -363,6 +363,10 @@ def test_pivot_sum(t):
     True  | 1        | 2        | 0
     """)
 
+def test_apply(t):
+    t = t.copy()
+    assert_array_equal(t.apply(lambda x, y: x * y, ['count', 'points']), np.array([9, 6, 6, 10]))
```
SamLau95: Let's add a test case here where we pass in only a string as the column_label.

SamLau95:
This PR could be better named :) and the description could be more informative. Ideally I'd be able to tell what this PR does and why we're doing this from the PR description/title itself.

SamLau95:
Great stuff! Updated the description of this PR a bit for posterity. Thanks for workin' on this.

PR #175, opened 2016-01-12
Added an apply_table function and test

choldgraf:
This is in response to [this gitter message](https://gitter.im/dsten/datascience?at=569437527669220736031fcc). It's a pretty simple implementation of an Apply method. It basically converts to a df, applies the function, and then converts back.

### DIFF
--- a/datascience/util.py
+++ b/datascience/util.py
@@ -1,11 +1,13 @@
 """Utility functions"""
 
-__all__ = ['percentile', 'plot_cdf_area']
+__all__ = ['percentile', 'plot_cdf_area', 'table_apply']
 
 
 import numpy as np
+import pandas as pd
 import matplotlib.pyplot as plt
 from scipy import stats
+
 
 def percentile(p, arr=None):
     """Returns the pth percentile of the input array (the value that is at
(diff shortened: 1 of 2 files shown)


datascience/util.py
```diff
@@ -53,3 +54,47 @@ def plot_cdf_area(rbound, lbound=None, mean=0, sd=1):
...
+        will be applied to all columns in table
+
+    Returns
+    -------
+    tab : instance of Table
+        A table with the given function applied. It will either be the
+        shape == shape(table), or shape (1, table.shape[1])
+    """
+    import pandas as pd
+    from datascience import Table
```
SamLau95: Could we move these imports up to the top with the rest of the imports?

choldgraf: Sure thing - was just trying to minimize unnecessary imports but if you
guys are cool with it I can pull 'em outside

choldgraf: Ah now I remember - I think I put the Table import in there because it was
raising import errors (presumably because of a circular import or
something). Nesting the Table import within the function makes it work
fine. But I put the pandas import at the top.

SamLau95: Yeah — I think what we decided is to wrap imports in a `try` block so we can still use readthedocs.

Also, we're going to be using `pandas` more for the `Table` class anyway so it doesn't benefit us very much to leave it here.

choldgraf: how does that work? just `try: from . import Table; except ImportError:
somethingsomething`? Anywhere in the code where you guys already do this so
I can just copy it?

SamLau95: Yeah, that's the basic idea. We aren't doing it yet because it was just discussed. It's not super high on the priority list at the moment but I expect it to happen in a few weeks as we set up versioning.

choldgraf: ok cool - well I've got the pandas import outside the function now, still have the Table import inside but I guess that'll be moved in a future PR. For now is there anything else you'd like to see change?

SamLau95: Nope! Thanks for the work on this.

cboettig:
yay! thanks.

PR #187, opened 2016-01-29
New Tables interface

papajohn:
I've changed the interface for table creation per our discussion a few weeks ago, but all old calls still run (with a deprecation warning). One breaking change is a different default function for `group` and `groups`. The original identity function was always problematic, because counts are usually desired.

I've incremented the version to indicate that there are breaking changes.


datascience/tables.py
```diff
@@ -182,62 +184,37 @@ def __getitem__(self, row_indices_or_slice):
...
         self._columns = collections.OrderedDict()
         self._formats = dict()
         self.formatter = formatter
 
-        # Ensure inputs are properly formed
-        columns = columns if columns is not None else []
-        labels = labels if labels is not None else []
-        assert len(labels) == len(columns), 'label/column number mismatch'
+        if _other is not None:
+            warnings.warn("deprecated", DeprecationWarning)
```
SamLau95: Looks like the warning doesn't show up (also tried in the repl):

screenshot 2016-01-28 23 21 30

papajohn: It's true that deprecation warnings are off by default in Python, and you
have to pass `-W` or something to get them to appear. I'm not sure we
should forcibly re-configure people's Python to show the warnings.

datascience/tables.py
```diff
@@ -406,15 +386,43 @@ def rows(self):
...
+
+        table.column(label) is equivalent to table[label].
+
+        Args:
+            label (int or str): The index or label of a column
+
+        Returns:
+            An array
+        """
+        return column[index_or_label]
```
SamLau95: Looks like this should be:

    return self._columns['x']

datascience/tables.py
```diff
@@ -674,8 +681,9 @@ def relabel(self, column_label, new_label):
 
     def copy(self):
         """Return a copy of a Table."""
+        # TODO(denero) Shallow copy by default with an option for deep copy
```
SamLau95: Does this TODO mean that it currently deep copies?

papajohn: Yes, I believe it deep copies because of the np.copy(self[label]), but I
don't think we really need to make deep copies ever. That seems like an
issue for another pull request, though.

datascience/tables.py
papajohn: Good point. Pushed a change that fixes `column` and deprecates `values`.

datascience/tables.py
SamLau95: Makes sense, although it seems likely that I wouldn't know a method was deprecated until it got removed entirely. At that point, I'd have breaking code.

SamLau95:
Ideally, the deprecation messages would say "use this method instead" and "this method will be deprecated in 0.6.0".

After we reach resolution on if / how we're going to make deprecation warnings appear this looks good to go.

Things off the top of my head to do either before/soon after merging:
- Record these changes in the changelog
- Update the tutorial with these changes
- Update the docs badge in the readme to point to this version

Are you still planning to release the rewrite with `pandas`?

datascience/tables.py
papajohn: FutureWarning looks like a good fit. It's supposed to indicate that
semantics will change in the future. It is shown by default. What do you
think?

papajohn:
Ok, I'm going to update the "deprecated" warnings to give direction about
improved versions. I think your list of other stuff is right, but I'd
rather save that for the weekend and push this now so that I can use it in
lecture & the textbook for tomorrow. It's unlikely that students will
really use the datascience package until lab next week, so we have a bit of
time to improve docs.

i think the pandas branch changes too much to be pushed out in a rush. I'd
rather push this branch, which makes the important interface changes
without breaking things along the way. Eventually I'd like to finish
pandas, but it's not as much of a priority as setting the right interface.

This new interface is consistent with what I've been building in the pandas
branch.

Thanks for your help tonight!

papajohn:
Pushed deprecation messages.

Issue #148, opened 2015-11-06
Table.hist should default to normed=True


davidwagner:
Ani says things in lecture like "in this class histograms are always normed".  So, let's make the default do the right thing: the default should be normed=True, so we don't need to repeat it (and students don't need to remember it).

a-adhikari:
Thank you! I think John suggested this a while ago too.
Ani

deculler:
Seems like this is an important question of terminology.  Doesn't this
choice make it a discrete distribution rather than a histogram.  There is a
related technical issue that comes up in boot strapping when we start
drawing samples from these whatever they are called.  In general a
histogram has bins.  Whereas when we start sampling we are drawing from a
discrete distribution, i.e, the bins are discrete values not ranges.  I
realize that I'm the nubee here but I would find this sort of inconsistency
very frustrating.

As I poke around now at the various definitions of histogram they do convey
discretization of a distribution.  So the counts per bin that everyone
learns in school and plots in excels a column chart is the aberration.  So
what is that thing called?

On Saturday, November 7, 2015, a-adhikari <notifications@github.com> wrote:

> Thank you! I think John suggested this a while ago too.
> Ani
>
> > On Nov 6, 2015, at 10:39 AM, davidwagner <notifications@github.com
> <javascript:_e(%7B%7D,'cvml','notifications@github.com');>> wrote:
> >
> > Ani says things in lecture like "in this class histograms are always
> normed". So, let's make the default do the right thing: the default should
> be normed=True, so we don't need to repeat it (and students don't need to
> remember it).
> >
> > —
> > Reply to this email directly or view it on GitHub.
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/dsten/datascience/issues/148#issuecomment-154510547>.
>


-- 
David E. Culler
Friesen Professor of Computer Science
Electrical Engineering and Computer Sciences
University of California, Berkeley

a-adhikari:
So the counts per bin that everyone
learns in school and plots in excels a column chart is the aberration. So
what is that thing called?
A mistake?

If anything, histograms turn discrete distributions into continuous ones:
mixtures of uniform distributions over the bins. The flat tops assume that
the proportion in the bin is uniformly smeared across the bin.

A histogram is essentially a density "curve", representing proportions by
areas. Nothing else is a histogram, to my knowledge. Those other things
might be bar graphs if you think of the bins as categorical.

Ani

[...]

Issue #137, opened 2015-10-22
Store plot history within a table


mDibyo:
In addition to making it easier to manipulate the plots, this will also make it easier to test the plotting functions. See discussion in #132.

SamLau95:
I actually am liking this idea more now because without this we can't test whether students have called hist properly during lab and projects.

papajohn:
It's too bad that the notebook will then display the object using `<...>` syntax before the visualization. We could instead store the history of plots as a field in the table. Thoughts?

Issue #185, opened 2016-01-24
Table.plot should take column_for_xticks as optional


deculler:
If it is not specified, it should be assumed to be 0:Table.num_rows.  Otherwise it is much harder to do the obvious thing of just plotting the values in a table.  And it is unnecessarily different from matplotlib plot.

cboettig:
:+1:  Likewise, if y values are listed explicitly, it would be nice if the `plot` method would plot only those y values and not other columns in the table.  Currently it is necessary to first subset the table, and then plot:

```py
tbl = Table([range(25), np.random.randn(25), np.random.randn(25), np.random.randn(25)], ["A", "B", "C", "D"])

# This plots A vs B, C, D, etc.., instead of A vs B.  That feels unexpected to me.  
tbl.plot("A", "B")

# meanwhile this feels redundant
tbl.select("A", "B").plot("A", "B")

```

As you probably know, `pandas` already has both the behavior David mentions and the behavior of plotting just the columns given in the `y` argument:

```py
## Plots just A vs B, 
tbl.to_df().plot("A", "B")

## Plots A vs B & C but not D
tbl.to_df().plot("A", ["B", "C"])

## Plots just B vs row number:
tbl.to_df().plot(y = "B")
```

deculler:
plot fails if column_for_xticks is unspecified.  This isn't picked up by the tests, as we don't have test for much of the plotting functions.  We could add tests for execution completion, even if we don't look at the quality of the graph.  A fairly comprehensive test is in the TableDemos notebook https://github.com/deculler/TableDemos/blob/master/BerkeleySalary.ipynb.  I'll PR the fix so this can be closed.

PR #193, opened 2016-02-07
Improve charts

papajohn:
Fix a bunch of issues with charts. Most significant changes are that (a) each charting method can select particular columns, and (b) `overlay` now defaults to `True`, because that's the option we always use.


papajohn:
Fixes #137 and #185

papajohn:
Also fixes #148

papajohn:
(and I think I'm done fiddling with this pull request.)

PR #195, opened 2016-02-10
Group docs

peterasujan:
Documentation update for select, drop, where, sort, group, and groups.


peterasujan:
``pivot`` and ``join`` are next!

datascience/tables.py
```diff
@@ -511,12 +511,12 @@ def copy(self, *, shallow=False):
...
     def select(self, column_label_or_labels):
-        """Return a Table with selected column or columns by label.
+        """Return a Table with selected column or columns by label or index.
 
         Args:
             ``column_label_or_labels`` (string or list of strings): The header
-            names of the columns to be selected. ``column_label_or_labels`` must
-            be an existing header name.
+            names or indices of the columns to be selected. ``column_label_or_labels`` must
+            be an existing header name, or a valid column index.
```
SamLau95: Good point with the index behavior. Could you add an example of that below?

datascience/tables.py
```diff
@@ -557,19 +557,56 @@ def exclude(self):
...
+
+        Args:
+            ``column_or_label``: The header name of a column in the table or an array.
+
+            ``value``: Value for comparison with items in ``column_or_label``.
+
+        Returns:
+            An instance of ``Table`` containing rows for which the ``column_or_label``
+            column or ``column_or_label`` itself is non-zero or True, or is equal to ``value``,
+            if provided.
```
SamLau95: Awesome! While we at it, could you write an example or two for this?

SamLau95:
Hooray!

PR #197, opened 2016-02-11
Fixing missing images in tutorial

peterasujan:


SamLau95:
Does this work locally? (eg. when you run `make docs` then visit `localhost`)

It looks like it's not working on the docs built for this branch :(

http://datascience.readthedocs.org/en/docs/tutorial.html

peterasujan:
Huh. That is a little weird. It's working here: http://peterasujan.github.io/datascience/tutorial.html (which should be equivalent to http://data8.org/datascience/tutorial.html), and locally.

peterasujan:
So things are starting to make a bit more sense. When I deployed the docs directly to gh-pages, it worked for a while, but then Travis ran and re-deployed from master (in response to John's push), overwriting the changes (which is why data8.org/datascience has reverted back). However, I'm confused why building based on the docs branch in readthedocs isn't working, unless its sphinx version is different or something.

SamLau95:
Getting the images to build properly was very finicky for me as well. Do the images still build properly if you delete the entire `_build` directory, then make the docs?

peterasujan:
Sorry for the slow response - was out of town for the weekend.
I removed both directories and rebuilt the docs, and it seems to work locally and when served from my [gh-pages](http://peterasujan.github.io/datascience/tutorial.html) branch (after I run ```make deploy_docs```).

Issue #184, opened 2016-01-22
Add warning when using read_table on a URL from data8/ds8


SamLau95:
In data-8/connector-instructors#13 Andrej ran into a cryptic error when trying to `read_table` on a `data8.berkeley.edu` link.

Since all of those links require authentication, none of them will work with `Table.read_table`. Chris suggested that we give up earlier with a more explicit message when we see one of those links. I think is a good idea as well.

deculler:
It is important that this not be specific to the course.  Can we clarify
what constitutes a path and what it is used for?

David E. Culler
Friesen Professor of Computer Science
Electrical Engineering and Computer Sciences
University of California, Berkeley

choldgraf:
I think the issue is that when you give a URL with the `data8` subdomain, then it passes to the authentication / login screen we usually get when we log in to data8. The `read_table` function doesn't know what to do with it, and so it throws a really cryptic error.

It sounds like for now doing this with the data8 domain will be the most common occurrence of this problem. If there were a way to detect that particular error (some kind of authentication error presumably) then maybe we could catch it and throw a more human-readable message. That might work for arbitrary domain names but I'm not sure how tractable it is to implement.

Another possibility would be to do a really simple try/catch statement, and for *any* error that gets thrown, just say `the URL you gave didn't work, please make sure it is correct and that it points to a data file readable by read_table` or something like that. Not the most elegant solution though.

papajohn:
I like the try/except solution, but the error message should give a hint to the probable cause, which is that the URL requires authentication.

choldgraf:
So I just looked at the `read_table` code and I'm not sure the in its current state we could do a simple try/except solution. This is because the code is already doing try/except. Here are the relevant lines (I think):

```python
        try:
            path = urllib.parse.urlparse(filepath_or_buffer).path
        except AttributeError:
            path = filepath_or_buffer
```

If it fails on the first try, then it assumes that it's working with a file and not a URL. So we'd have to add an except statement for the specific error that gets thrown with an authentication problem. Alternatively, we could add a check in the first try statement, something like:

```python
        try:
            path = urllib.parse.urlparse(filepath_or_buffer).path
            if 'data8.berkeley.edu' in filepath_or_buffer:
                raise ValueError('data8 filepaths require authentication, host your data on a public server')
        except AttributeError:
            path = filepath_or_buffer
```

Thoughts?

SamLau95:
I think this is fine, personally speaking, because:
[...]

PR #198, opened 2016-02-13
Throwing an error when data8.berkeley.edu URL is passed

choldgraf:
Throws a warning when `read_table` gets a URL with data8.berkeley.edu in it. Otherwise it will break w/ an unhelpful error because of authentication problems. See #184

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -2,7 +2,6 @@ import doctest
 import re
 import pytest
 import numpy as np
-from nose.tools import assert_raises
 from numpy.testing import assert_array_equal
 from datascience import *
 import pandas as pd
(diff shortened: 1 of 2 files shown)


tests/test_tables.py
```diff
@@ -1038,3 +1039,9 @@ def test_array_roundtrip(table):
 
     for (c0, c1) in zip(t.columns, table.columns):
         assert_equal(c0, c1)
+
+
+def test_url_parse():
+    """Test that Tables parses URLs correctly"""
+    url = 'https://data8.berkeley.edu/something/something/dark/side'
+    assert_raises(ValueError, Table.read_table, url)
```
SamLau95: Looks like we can just do

```python
with pytest.raises(ValueError):
    Table.read_table(url)
```

So we don't have to require `nose`.

choldgraf:
ah ya good point, I should have looked into how you guys were doing it before

SamLau95:
Thanks for doing this!

PR #199, opened 2016-02-14
Making xticks work when rotated

choldgraf:
Currently if xticks are > length 5 then they're incorrectly rotated, making them have the wrong values.


papajohn:
Fixed in another branch and merged.

Issue #207, opened 2016-02-24
Extend where method functionality


papajohn:
Proposal: create a plain-language interface for filtering based on one column. E.g., assuming `x` is a label in `t`

```
t.where_column('x', equals(3))
t.where_column('x', between(-2, 2))
t.where_column('x', less_than(4))
t.where_column('x', not less_than(4))
t.where_column('x', less_than(4) or greater_than(6))
```

We could even let functions on the original column be arguments to the comparators.

```
t.where_column('x', equals(np.mean))
t.where_column('x', not equals(max))
```

Thoughts?

SamLau95:
I actually think we could add this functionality to `where`, since it already supports statements like:

    t.where('x', 2)

We could add the ability to pass in a function as the second argument, which would look like your examples above.

I like this because it solves a common problem with chaining `where` calls:

    t.where(t['x'] != 2).where(t['y'] == 4) # doesn't work
    t.where('x', not_equals(2)).where('y', equals(4)) # works

papajohn:
I like it. `t.where('x', between(2, -2))` would save a lot of typing.

PR #209, opened 2016-03-05
Predicates (and fix minimize so that tests pass)

papajohn:


papajohn:
Implements #207

datascience/__init__.py
```diff
@@ -3,4 +3,5 @@
 from .tables import *
 from .formats import *
 from .maps import *
+from .predicates import *
```
SamLau95: This means that all the functions in `predicates.py` will be in the global scope when we run `from datascience import *`.

Are we concerned about potential name conflicts since `predicates.py` has a lot of functions?

datascience/predicates.py
```diff
@@ -0,0 +1,63 @@
+"""Predicate functions."""
```
SamLau95: I think it'd be a good idea to also define the `__all__` variable for this file.

SamLau95:
Looks mostly good to me, aside from a question and a suggestion.

We should also add `predicates.py` to the docs by adding a line in `docs/index.rst` and creating a `predicates.rst` file that looks like the `util.rst` file.

papajohn:
I suppose we could put all predicate functions as attributes of some
class... I'll think about it. Something like,

t.where('age', Item.between(3, 8))

peterasujan:
If you end up pursuing this method, I might suggest using "is" (I guess it should be capitalized though) as the class name, so that the whole line reads (almost) like English.

papajohn:
Ok, all single predicates are now methods of `are`. (I considered `Is`, but requires caps to avoid keyword and `ls` and `Is` look the same.)

I left `both` and `either` in the global namespace as well as in `are` because it looks prettier... 

```
t.where('age', either(are.equal_to(5), are.above(8))
```
vs
```
t.where('age', are.either(are.equal_to(5), are.above(8))
```

PR #225, opened 2016-06-27
Reduce use of lists

### LINKED ISSUE #224: Reduce use of lists in a few Table methods
A few table methods take lists as arguments when they could be varargs methods.  This seems to make the learning curve slightly steeper for some students.

We'll also add a function to make an array from a list of arguments (rather than from a list), like make_array(1, 2, 3).

henryem:
Resolves #224.


SamLau95:
These look good to me if @papajohn is okay with these changes!

Thanks for updating changelog and bumping version :)

papajohn:
LGTM!

Issue #226, opened 2016-07-07
as_array returns an (n x 1) array


choldgraf:
Currently using a table's `to_array` returns an array of shape `n x 1`. It seems more intuitive that this function would return an `n x n_columns` array instead (at least, this has confused several of the people I've been developing a course with). I think this would be an easy fix, but perhaps there's a reason that it's returning `n x 1` so I thought I'd ask before making a PR. Would people be open to a PR that would do something like this instead?

```python
# Determine output data type (object if multiple types)
dtypes = [c.dtype for c in xy.columns]
dt_out = np.object if len(set(dtypes)) != 1 else dtypes[0]

# Populate output array
arr = np.empty([xy.num_rows, xy.num_columns], dt_out)
for i, col in enumerate(xy.columns):
    arr[:, i] = col
return arr
```

SamLau95:
This sounds fine to me.

In the textbook:

    grep to_array notebooks -r | wc -l
    0

In the `data8assets` repo:

    grep to_array . -r | wc -l
    0

Looks like we don't it in the foundations class.

choldgraf:
oh? how do you get the data in a `Table` as a numpy array?

SamLau95:
I think we typically do something like `table['latitude'][5]` to get the element at index 5 of the column `latitude` in the table. You can also do `table[4][5]` to get the element in column 4 (0-indexed), row 5.

choldgraf:
So there's no way to just turn the whole thing into an array? (e.g., if we
want to pass the whole Table to a function that only accepts Numpy arrays)

SamLau95:
Looks like there isn't; sounds like `to_array` should do that.

choldgraf:
I see that there's a test in there is a test in there:

`test_array_roundtrip(table)`

that assumes that the output of `to_array` will be a structured array instead of an ndarray. So maybe somebody assumed that this is the desired behavior?

As far as getting the values of the data as an array, right now these are the (kind of hacky) options:

1. np.array(table.rows)
1. table.to_df().values

So it seems like these would be easy ways to do the array conversion. Perhaps `to_array` could have a `structured=False` parameter that would define whether to make it a structured array or as a simple ndarray?

WDYT?

SamLau95:
I believe this is a function that @stefanv added about a year ago — maybe he has more input? It seems like this function is used infrequently enough (if at all) that I'm fine with either changing the output or adding a param.

stefanv:
[...]

PR #227, opened 2016-07-19
making .values return a numpy array

choldgraf:
this PR makes `table.values` return a numpy array (since it was being deprecated and isn't used in the textbook). This makes its behavior roughly equivalent to pandas dataframes. addresses #226 @SamLau95 @stefanv

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -266,11 +266,20 @@ class Table(collections.abc.MutableMapping):
         """
         return self._columns[self._as_label(index_or_label)]
 
-    # Deprecated
-    def values(self, label):
-        """Returns the values of a column as an array. [Deprecated]"""
-        warnings.warn("values is deprecated; use column", FutureWarning)
-        return self.column(label)
+    @property
+    def values(self):
+        """Return data in `self` as a numpy array.
+
+        If all columns are the same dtype, the resulting array
+        will have this dtype. If there are >1 dtypes in columns,
+        then the resulting array will have dtype `object`.
+        """
+        dtypes = [col.dtype for col in self.columns]
+        if len(set(dtypes)) > 1:
+            dtype = object
+        else:
+            dtype = None
+        return np.array(self.rows, dtype=dtype)
 
     def column_index(self, column_label):
         """Return the index of a column."""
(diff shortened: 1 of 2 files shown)


choldgraf:
I could change this to call `table.to_df().values` instead, though figured I'd try it with pure numpy first.

datascience/tables.py
```diff
@@ -266,12 +266,16 @@ def column(self, index_or_label):
         """
         return self._columns[self._as_label(index_or_label)]
 
-    # Deprecated
-    def values(self, label):
-        """Returns the values of a column as an array. [Deprecated]"""
-        warnings.warn("values is deprecated; use column", FutureWarning)
-        return self.column(label)
+    @property
+    def values(self):
```
SamLau95: Think you could add a docstring + a test for this function? :)

SamLau95:
cc'ing @papajohn . This LGTM after adding docstring + test. Oh, and could you also add a note in the CHANGELOG.md?

(Might also want to bump the version number if you need to `pip install` this right away.)

choldgraf:
Ah looks like this'll break a test: L493 of test_tables.py  (this tests the deprecated `values` function). Do you guys think it's OK to change this test?

choldgraf:
(and not a big rush on this)

datascience/tables.py
```diff
@@ -266,12 +266,16 @@ def column(self, index_or_label):
...
+        dtypes = [col.dtype for col in self.columns]
+        if len(set(dtypes)) > 1:
+            dtype = object
+        else:
+            dtype = None
+        return np.array(self.rows, dtype=dtype)
 
+    # Deprecated
```
papajohn: Is this supposed to be deprecated, or did the comment just move from the old values function?

papajohn:
I have no problem with replacing the old `.values()`. Documentation would be great, as @SamLau95  suggested.

datascience/tables.py
choldgraf: just moved from the old values function...I assume it was a section header?

choldgraf:
Sounds good - I'll add a docstring + test

choldgraf:
Also, looks like I'm an idiot and mistook a dictionary for a table 😆 ...I don't think this will break a test after all but I'll add a new one regardless

datascience/tables.py
papajohn: Please remove the comment. `values` was deprecated, but `column_index` is not. Thanks.

choldgraf:
Mmk, added a test for this as well as a docstring. Also removed the `# Deprecated` comment because I think it was attached only to the deprecated `values()` function after all

stefanv:
Looks good to me.

SamLau95:
LGTM too! 🚢

choldgraf:
hooray!

Issue #231, opened 2016-07-23
Add some predicates for strings


henryem:
I'd like to add a few predicates that are specialized for strings, like `are.containing`.  It will make the predicate API a little more elaborate, which is unfortunate.  Since we're going to try hiding booleans from students for the first month+, I think the flexibility we gain in working with text is worth the cost.

Use case: Right now I'm trying to write an exercise about news articles, and I'd like to find all the articles whose titles contain the word "Ecuador".  The best workaround right now is something like 

`def ecuador_location(string): return string.find("Ecuador")`
`tbl.with_column("Ecuador location", tbl.apply(ecuador_location, "Title").where("Ecuador location", are.above(-1))`

...which is pretty ugly, and not really easier to understand than boolean values.

PR #232, opened 2016-07-24
Added some predicates for strings.

henryem:
[\*] Wrote test for feature
[\*] Added note about PR in CHANGELOG.md
[\*] Bumped version number (delete if unneeded)

**Changes proposed:**
Added two predicates for strings in tables: `containing` and `contained_in`.  See #231 .


SamLau95:
This looks good to me!

henryem:
@SamLau95 can you push this version (0.7.0) to the pypi repo?  We also need to upgrade the data8/ds8 servers.

SamLau95:
@ryanlovett I pushed the version to pypi and updated the systemuser Dockerfile. If you're updating the JHub proxy tonight think you can push this change through too? (I didn't push a new container to DockerHub.)

ryanlovett:
Yes, will do.

Ryan, from my phone

ryanlovett:
Fyi, docker hub is setup to automatically rebuild data8/systemuser when the
Dockerfile changes in the git repo. I'm pulling to the nodes now.


Ryan

ryanlovett:
@SamLau95 I just noticed that you committed the 0.7.0 change to the systemuser master branch whereas the deployments are using the nodrive tag. I just committed the same change to :nodrive and will pull to the nodes after it is built.

SamLau95:
Oops, thanks for the catch!


On September 8, 2016 at 4:24:05 PM, Ryan Lovett (notifications@github.com) wrote:

@SamLau95 I just noticed that you committed the 0.7.0 change to the systemuser master branch whereas the deployments are using the nodrive tag. I just committed the same change to :nodrive and will pull to the nodes after it is built.

henryem:
Can you let us know when the change is available?  A student (one of the few who actually run the textbook examples, apparently) noticed recently that a textbook example that requires 0.7.0 doesn't work.

ryanlovett:
Pulling right now... should be about 5-10 minutes after which students
would need to stop and start their server to get the new image.

PR #237, opened 2016-09-12
Under-the-hood changes for table.values

choldgraf:
[ ] Wrote test for feature
[ ] Added note about PR in CHANGELOG.md

**Changes proposed:**
This is a backend change to the Table.values property. Somewhere recently the behavior of `table.rows` changed to return a `Rows` object instead of a list of arrays (which is what `table.columns` does). This was taking a looong time when passed to `np.array`, so I'm changing it to use `table.columns` instead. We should think of a different solution if the plan is to change the output of `table.columns` as well.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -279,7 +279,7 @@ class Table(collections.abc.MutableMapping):
             dtype = object
         else:
             dtype = None
-        return np.array(self.rows, dtype=dtype)
+        return np.array(self.columns, dtype=dtype).T
 
     def column_index(self, column_label):
         """Return the index of a column."""


SamLau95:
This looks good to me! If you need this right away, think you can bump the version number and update the changelog?

choldgraf:
I'm not sure if I need it right now or not. I've been trying to use `datascience` for my lectures (e.g. instead of pandas) but now that I'm running my notebooks on the jupyterhub instance, I'm getting a bunch of performance errors (e.g. I don't think `datascience` is very memory efficient, because I keep crashing the kernel). If that's the case then I might have to just switch to use `pandas`.

choldgraf:
so after some digging it seems like not all `Table` creation methods were created equally...if I use `with_columns` then the kernel dies (I think because of memory). If I convert the data to a DataFrame first, then use `from_df` then it works. Is there a guide for recommended ways to create `table` instances?

SamLau95:
cc @papajohn

Darn, I see. Think you could send us a notebook that’s crashing? An interact link works too.

We usually use Table.read_table (which calls the pandas read_table method) to read in data (eg. Table.read_table('imdb.csv')). Not sure if that’s the most memory-efficient way though.

We haven’t run into memory issues in Data 8 so we’re heading into uncharted territory for datascience.

John was working on implementing Table as a thin layer on top of a pandas dataframe. I wonder what the status of that is, since that would make datascience and pandas basically equivalent from a memory / performance standpoint.



--
Sam Lau
EECS | UC Berkeley
Project Mentor | Blueprint, Technology for Nonprofits
samlau95@gmail.com

On September 11, 2016 at 6:20:24 PM, Chris Holdgraf (notifications@github.com) wrote:

so after some digging it seems like not all Table creation methods were created equally...if I use with_columns then the kernel dies (I think because of memory). If I convert the data to a DataFrame first, then use from_df then it works. Is there a guide for recommended ways to create table instances?

choldgraf:
yeah, so neuroscience data is notoriously crappy for memory purposes. We have a small-ish dataset, and it's currently at ~200k rows by 60 columns. So I can make my kernel crash by just doing `data = np.random.randn(300000, 50); ds.Table().with_columns([(str(ii), idata) for ii, idata in enumerate(data.T)`

re: using a thin layer around a dataframe, I think that's a great idea. Pandas has put a ton of time into efficiency etc, so avoiding reinventing the wheel if it doesn't result in really weird APIs sounds great to me. In fact a lot of my code is just turning a table into a dataframe, then doing a bunch of stuff in pandas, then turning it back into a table.

papajohn:
I still like the idea of implementing tables as data frames, but not this
semester.

PR #241, opened 2016-09-26
v0.8.0: Change table.sample behavior and add Map.overlay

### LINKED ISSUE #236: Mapping enhancements
@papajohn @SamLau95 
Hi! I'm the GSI for the smart cities connector course. I talked to Sam about the mapping tool this morning. I was wondering if we could add a method to overlay map features over an existing map. Basically the issue I ran into is that after I create a map I may want to add markers or labels to the map, and so far there's not an easy way to do that:

I wrote the following method that seems to do the trick. Could this be added to the Map tool?
The only other issue with the function is that in my implementation I edit the Map._folium_map directly. It might be nice if we copy and return a new Map object instead of modifying the existing Map in place. I couldn't figure out the best way to do this. If you have time, it would be great if you could take a look. Let me know if you have any questions.

Best,
Maddie

def overlay(feature, ds_map, color='Blue', opacity=.6):
    '''
    feature can be a table of map features, a list of map features,
    a Map, or a Region. The features will be drawn on the existing 
    Map (ds_map). 
    '''
        
    if type(feature) == Table:
        # if table of features e.g. Table.from_records(taz_map.features)
        if 'feature' in feature:
            feature = feature['feature']
            
        # if marker table e.g. table with columns: latitudes,longitudes,popup,color,radius
        else:
            feature = Circle.map_table(feature)
            
    if type(feature) in [list, np.ndarray]:
        for f in feature:
            f._attrs['fill_color'] = color
            f._attrs['fill_opacity'] = opacity
            f.draw_on(ds_map._folium_map)
    
    elif type(feature) == Map:
        for i in range(len(feature._features)):
            f = feature._features[i]
            f._attrs['fill_color'] = color
            f._attrs['fill_opacity'] = opacity
            f.draw_on(ds_map._folium_map)
    elif type(feature) == Region:
        feature._attrs['fill_color'] = color
        feature._attrs['fill_opacity'] = opacity
        feature.draw_on(ds_map._folium_map)
    return ds_map

SamLau95:
**Breaking changes**

- Change default behavior of `table.sample` to `with_replacement=True` instead
  of `False`. (3717b67)

**Additions**

- Added `Map.copy`.
- Added `Map.overlay` which overlays a feature(s) on a new copy of Map.
  (315bb63e)

Closes #236


SamLau95:
Merging now since both Ani and Maddy need these changes tomorrow.

PR #245, opened 2016-10-01
With columns

maxwelljweinstein:
with_columns docs updated...

@SamLau95


datascience/tables.py
```diff
@@ -1328,98 +1328,110 @@ def with_rows(self, rows):
...
-        ... ])
-        >>> tiles.with_column('points', [3, 2])
-        letter | count | points
-        c      | 2     | 3
-        d      | 4     | 2
-        >>> tiles.with_column('count', 1)
+        Returns:
+            copy of original table with new or replaced column
+
+        >>> alphabet
```
SamLau95: The python code in the docstring actually gets run as doctests (https://docs.python.org/3/library/doctest.html), so this line makes the build fail (https://travis-ci.org/data-8/datascience/builds/164195943) since the `alphabet` table isn't defined.

You'll want to create it using `with_columns` first.

datascience/tables.py
```diff
@@ -1328,98 +1328,110 @@ def with_rows(self, rows):
...
+                the existing column will be replaced in the new table.
 
             ``values`` (single value or sequence): If a single value, every
-                value in the new column is ``values``.
+                value in the new column is ``values``. If sequence of values, 
+                new column takes on values in ``values``. 
 
                 If a sequence, the new column contains the values in
-                ``values``. ``values`` must be the same length as the table.
+                ``values``, must be the same length as the table.
```
SamLau95: Isn't this kind of a repeat of the last line but with the length constraint? Feels like you should just mention the length constraint here.

datascience/tables.py
```diff
@@ -1328,98 +1328,110 @@ def with_rows(self, rows):
...
+            
 
         Args:
-            ``labels_and_values``: An alternating list of labels and values or
-                a list of label-values pairs.
+            ``labels_and_values``: An alternating list of labels and values or a list of label-values pairs. 
+            Values can be any combination of sequences and single values. If one of the labels is in existing table, 
+            the existing column is replaced by its values pair. If any of the values is just a single value (``int``), 
+            every value in the corresponding column is that value. Columns arranged in order of labels as they appear in 
+            ``labels_and_values``.
```
SamLau95: nit: Please keep lines to under 79 characters (https://www.python.org/dev/peps/pep-0008/#maximum-line-length). Also, could you indent this text one level like the method you changed above?

Let's just take out the "Values can be any combination of sequences and single values." since it's confusing.

> If one of the labels is in existing table, the existing column is replaced by its values ~~pair~~.

How about:
> ...every value in the corresponding column is set to that value.

And:
> Columns added in order of labels

docs/index.rst
```diff
@@ -5,7 +5,7 @@ Welcome to datascience's documentation!
 :Date: |today|
 
 The ``datascience`` package was written for use in Berkeley's DS 8 course and
-contains useful functionality for investigating and graphically displaying data.
+contains useful functionality for investigating and graphically displaying data. Hello Sam.
```
SamLau95: Looks like this is leftover from earlier today!

datascience/tables.py
```diff
@@ -1332,30 +1332,29 @@ def with_column(self, label, values):
 
         Args:
             ``label`` (str): The column label. If an existing label is used,
-                the existing column will be replaced in the new table.
+            the existing column will be replaced in the new table.
```
SamLau95: Oh, but this should be reindented.

maxwelljweinstein: No idea why it's complaining that the table isn't defined in the doc test... any ideas?

maxwelljweinstein: Ok, issue is that errors of the throws, i.e tracebacks to the assertions errors, aren't exactly equivalent... (I don't print the entire error message, only a snip-it). It should be fixed except for those cases now.

SamLau95:
Looks like the build is still failing. If you click the Details link to the side it'll bring you to a page showing the errors:

```
=================================== FAILURES ===================================
[1m[31m________________________________ test_doctests _________________________________[0m

[1m    def test_doctests():[0m
[1m        results = doctest.testmod(tables)[0m
[1m>       assert results.failed == 0[0m
[1m[31mE       assert 17 == 0[0m
[1m[31mE        +  where 17 = TestResults(failed=17, attempted=142).failed[0m

[1m[31mtests/test_tables.py[0m:90: AssertionError
----------------------------- Captured stdout call -----------------------------
**********************************************************************
File "/home/travis/build/data-8/datascience/datascience/tables.py", line 1351, in datascience.tables.Table.with_column
Failed example:
    alphabet = Table().with_column('letter', make_array('c','d'))
Exception raised:
    Traceback (most recent call last):
      File "/home/travis/miniconda/envs/datascience/lib/python3.4/doctest.py", line 1318, in __run
        compileflags, 1), test.globs)
      File "<doctest datascience.tables.Table.with_column[0]>", line 1, in <module>
        alphabet = Table().with_column('letter', make_array('c','d'))
    NameError: name 'make_array' is not defined
**********************************************************************
....
```

`make_array` isn't defined so it looks like you have an old version of the `datascience` code.

Can you run:

```
git checkout master
git pull origin master # pulls from master to get latest code
git checkout with_columns
git merge master # pulls code into your branch
```

And push again? I think it'll work after that.

SamLau95:
Awesome! Can you merge this PR and watch Travis (https://travis-ci.org/data-8/datascience) for the build status? Once the build completes, you should verify that the docs are updated on http://data8.org/datascience

maxwelljweinstein:
All set, build passed, docs updated... onto more!
ᐧ

PR #247, opened 2016-10-05
take method docs updated

maxwelljweinstein:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -2340,20 +2340,25 @@ class _RowSelector(metaclass=abc.ABCMeta):
 
 class _RowTaker(_RowSelector):
     def __getitem__(self, row_indices_or_slice):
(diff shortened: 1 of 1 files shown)


datascience/tables.py
```diff
@@ -2340,65 +2340,52 @@ def __getitem__(self, item):
...
         A+           | 4
-        >>> t.take(5)
-        letter grade | gpa
-        B-           | 2.7
-        >>> t.take(-1)
+        >>> grades.take(-1)
         letter grade | gpa
         B-           | 2.7
-        >>> t.take([2, 1, 0])
+        >>> grades.take(make_array(2, 1, 0))
```
SamLau95: Think you could also show an example with `np.arange`? That's a common case.

PR #252, opened 2016-10-22
Sample pivot

maxwelljweinstein:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -31,7 +31,7 @@ class Table(collections.abc.MutableMapping):
     def __init__(self, labels=None, _deprecated=None, *, formatter=_formats.default_formatter):
         """Create an empty table with column labels.
 
-        >>> tiles = Table(['letter', 'count', 'points'])
+        >>> tiles = Table(make_array('letter', 'count', 'points'))
         >>> tiles
         letter | count | points
 
@@ -324,10 +324,10 @@ class Table(collections.abc.MutableMapping):
             A numpy array consisting of results of applying ``fn`` to elements
             specified by ``column_label`` in each row.
 
-        >>> t = Table().with_columns([
-        ...     'letter', ['a', 'b', 'c', 'z'],
-        ...     'count',  [9, 3, 3, 1],
-        ...     'points', [1, 2, 2, 10]])
+        >>> t = Table().with_columns(
+        ...     'letter', make_array('a', 'b', 'c', 'z'),
+        ...     'count',  make_array(9, 3, 3, 1),
+        ...     'points', make_array(1, 2, 2, 10))
         >>> t
         letter | count | points
         a      | 9     | 1
@@ -403,8 +403,8 @@ class Table(collections.abc.MutableMapping):
         """Appends a column to the table or replaces a column.
 
         ``__setitem__`` is aliased to this method:
-            ``table.append_column('new_col', [1, 2, 3])`` is equivalent to
-            ``table['new_col'] = [1, 2, 3]``.
+            ``table.append_column('new_col', make_array(1, 2, 3))`` is equivalent to
+            ``table['new_col'] = make_array(1, 2, 3)``.
 
         Args:
             ``label`` (str): The label of the new column.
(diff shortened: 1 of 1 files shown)


maxwelljweinstein:
@SamLau95 Want to look this over before I do the manual merge?

SamLau95:
Can you do the merge from master but not merge in the pull request? Eg. run `git pull origin master` on this branch and `git push origin sample_pivot` but don't click Merge pull request.

datascience/tables.py
```diff
@@ -1003,15 +1004,58 @@ def groups(self, labels, collect=None):
...
+        """Generate a table with a column for each unique value in ``columns``,
+        with rows for each unique value in ``rows``. Each row counts/aggregates
+        the values that match both row and column based on ``collect``.
+
+        Args:
+            ``columns`` -- a single column label, (``str``), in self, used to
+                create new columns, based on its unique values in self.
+            ``rows`` -- row labels, as (``str``) or list of strings, used to
+                create new rows based on it's unique values.
+            ``values`` -- column label in self for use in aggregation.
```
SamLau95: For these arguments, can we avoid the use of `self`? The students don't know what that means. We should instead say `the table`.

datascience/tables.py
```diff
@@ -1003,15 +1004,58 @@ def groups(self, labels, collect=None):
...
-        collect -- aggregation function over values
-        zero -- zero value for non-existent row-column combinations
+        """Generate a table with a column for each unique value in ``columns``,
+        with rows for each unique value in ``rows``. Each row counts/aggregates
+        the values that match both row and column based on ``collect``.
+
+        Args:
+            ``columns`` -- a single column label, (``str``), in self, used to
+                create new columns, based on its unique values in self.
+            ``rows`` -- row labels, as (``str``) or list of strings, used to
```
SamLau95: nit: Use `array` instead of `list`.

datascience/tables.py
```diff
@@ -1202,23 +1246,21 @@ def sample(self, k=None, with_replacement=True, weights=None):
...
-            k (int or None): If None (default), all the rows in the table are
-                sampled. If an integer, k rows from the original table are
-                sampled.
+        Args:
+            ``k`` -- specifies the number of rows (``int``)  to be sampled from
+                self. Default is k is equal to number of rows in self.
 
-            with_replacement (bool): If True (default), samples the rows with
-                replacement. If False, samples the rows without replacement.
+            ``with_replacement`` -- (``boolean``), if true samples ``k`` rows
```
SamLau95: nits: Technically there's only `bool`s in Python, not `boolean`s. Also, could you capitalize `true` and specify which option is the default? Finally, the same comment about using `self` applies here.

datascience/tables.py
```diff
@@ -1202,23 +1246,21 @@ def sample(self, k=None, with_replacement=True, weights=None):
...
-            weights (list/array or None): If None (default), samples the rows
-                using a uniform random distribution. If a list/array is passed
-                in, it must be the same length as the number of rows in the
-                table and the values must sum to 1. The rows will then be
-                sampled according the the probability distribution in
-                ``weights``.
+            ``weights``: Array specifying valid probability distribution.
+                Rows in self are sampled according the the
+                probability distribution given by ``weights``. Default is
+                uniform distribution on [1, ... , n], n = number of rows.
```
SamLau95: I actually think the previous description is clearer, since it specifies the format of `weights`. If you feel like yours is better, care to explain what you didn't like about the previous description?

datascience/tables.py
```diff
@@ -1250,6 +1292,16 @@ def sample(self, k=None, with_replacement=True, weights=None):
...
+        >>> jobs.sample(k = 2, weights = make_array(1, 0, 1, 0))
+        Traceback (most recent call last):
+            ...
+        ValueError: probabilities do not sum to 1
+
+        # Weights must be length of table.
+        >>> jobs.sample(k = 2, weights = make_array(1, 0, 0))
+        Traceback (most recent call last):
+            ...
+        ValueError: a and p must have same size
```
SamLau95: Could you add a `Raises` section to the docstring to document these `ValueError`s?

datascience/tables.py
```diff
@@ -1247,20 +1260,25 @@ def sample(self, k=None, with_replacement=True, weights=None):
...
         Args:
-            ``k`` -- specifies the number of rows (``int``)  to be sampled from
-                self. Default is k is equal to number of rows in self.
+            ``k`` -- specifies the number of rows (``int``) to be sampled from
+               the table. Default is k equal to number of rows in the table.
 
-            ``with_replacement`` -- (``boolean``), if true samples ``k`` rows
-                with replacement from self, else samples ``k`` rows without
-                replacement.
+            ``with_replacement`` -- (``bool``) By default TRUE; Samples ``k``
```
SamLau95: Oops, I was unclear in my previous comment. I meant `True`, since that's how boolean values in Python are spelled.

datascience/tables.py
```diff
@@ -1202,23 +1259,26 @@ def sample(self, k=None, with_replacement=True, weights=None):
...
+
+            ``with_replacement`` -- (``bool``) By default TRUE; Samples ``k``
+                rows with replacement from table, else samples ``k`` rows
+                without replacement.
 
-            with_replacement (bool): If True (default), samples the rows with
-                replacement. If False, samples the rows without replacement.
+            ``weights`` -- Array specifying probability the ith row of the
+                table is sampled. If None, by default, ``weights`` is the
+                uniform distribution on [1, ... , n], n = number of rows.
```
SamLau95: I don't think this is more clear than before, which states:

    If None (default), samples the rows using a uniform random distribution.

maxwelljweinstein: I don't think it was clear how the sampling was done... specifically we are sampling indices and then selecting. That's why I wanted to make explicit the ith entry in the array of weights corresponds to the probability that the ith row is selected, in sample size 1.

maxwelljweinstein:
@SamLau95  all set?

SamLau95:
Yup, I already approved it so you can merge whenever you'd like.

PR #253, opened 2016-10-31
Relabel/Relabeled/Apply

maxwelljweinstein:
Updated docs

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -306,23 +306,24 @@ class Table(collections.abc.MutableMapping):
         return self.labels.index(column_label)
 
     def apply(self, fn, column_label=None):
-        """Returns an array where ``fn`` is applied to each set of elements
-        by row from the specified columns in ``column_label``. If no
-        column_label is specified, then each row is passed to fn.
+        """ Apply ``fn`` to each element of ``column_label``.
+        If no ``column_label`` provided, `fn`` applied to each row of table.
 
         Args:
-            ``fn`` (function): The function to be applied to elements specified
-                by ``column_label``.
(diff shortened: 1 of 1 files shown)


maxwelljweinstein:
@SamLau95

datascience/tables.py
```diff
@@ -1578,26 +1587,48 @@ def with_columns(self, *labels_and_values):
...
 
-            ``new_label`` (str or sequence of str): The new label(s) of
+            ``new_label`` -- (str or array of str): The new label(s) of
                 columns to be changed. Same number of elements as label.
 
-        >>> tiles = Table(['letter', 'count'])
-        >>> tiles = tiles.with_rows(
-        ...    make_array(make_array('c', 2), make_array('d', 4)))
+        Raises:
+            ``ValueError`` -- if ``label`` or ``new_label`` do not exist in
```
SamLau95: nit: `new_label` doesn't have to exist in the original table, only `label`.

datascience/tables.py
```diff
@@ -1578,26 +1587,48 @@ def with_columns(self, *labels_and_values):
...
+        >>> tiles
+        letter | count
+        c      | 2
+        d      | 4
+        >>> tiles.relabeled(make_array('letter', 'count'),
+        ...   make_array('column1', 'column2'))
+        column1 | column2
+        c       | 2
+        d       | 4
+        >>> tiles.relabeled(make_array('letter', 'number'),
```
SamLau95: Think you could add a comment here saying that this example shows that the original table is unmodified by `relabeled`?

PR #254, opened 2016-11-27
Join, Barh, Scatter, Plot docs

maxwelljweinstein:
Join docs created...
Barh, plot, scatter docs all updated and uniform.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1870,12 +1870,6 @@ class Table(collections.abc.MutableMapping):
     def plot(self, column_for_xticks=None, select=None, overlay=True, **vargs):
         """Plot line charts for the table.
 
-        Each plot is labeled using the values in `column_for_xticks` and one
-        plot is produced for all other columns in self (or for the columns designated
-        by `select`).
-
-        Every selected column except for `column_for_xticks` must be numerical.
-
         Args:
             column_for_xticks (``str/array``): A column containing x-axis labels
 
@@ -1886,6 +1880,13 @@ class Table(collections.abc.MutableMapping):
             vargs: Additional arguments that get passed into `plt.plot`.
                 See http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.plot
                 for additional arguments that can be passed into vargs.
+        Raises:
+            ValueError -- Every selected column must be numerical.
+
+        Returns:
+            Returns a line plot (connected scatter). Each plot is labeled using
+            the values in `column_for_xticks` and one plot is produced for all
+            other columns in self (or for the columns designated by `select`).
         >>> table = Table().with_columns(
         ...     'days',  make_array(0, 1, 2, 3, 4, 5),
         ...     'price', make_array(90.5, 90.00, 83.00, 95.50, 82.00, 82.00),
@@ -1972,12 +1973,6 @@ class Table(collections.abc.MutableMapping):
     def barh(self, column_for_categories=None, select=None, overlay=True, **vargs):
         """Plot horizontal bar charts for the table.
 
-        Each plot is labeled using the values in ``column_for_categories`` and
-        one plot is produced for every other column (or for the columns
-        designated by ``select``).
-
-        Every selected except column for ``column_for_categories`` must be numerical.
-
         Args:
             ``column_for_categories`` (``str``): A column containing y-axis categories
                 used to create buckets for bar chart.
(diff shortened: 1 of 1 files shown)


maxwelljweinstein:
@SamLau95

datascience/tables.py
```diff
@@ -1155,12 +1155,54 @@ def stack(self, key, labels=None):
         return type(self)([key, 'column', 'value']).with_rows(rows)
 
     def join(self, column_label, other, other_label=None):
-        """Generate a table with the columns of self and other, containing rows
+        """Generates a new table with the columns of self and other, containing rows
```
SamLau95: nit: How about just "Create a new table"? I think that's more consistent with the language we use in other methods.

datascience/tables.py
```diff
@@ -1155,12 +1155,54 @@ def stack(self, key, labels=None):
...
         for all values of a column that appear in both tables.
-        If a join value appears more than once in self, each row will be used,
-        but in the other table, only the first of each will be used.
```
SamLau95: I think this comment is an important implementation detail! But right now it's just worded weirdly. Basically, the behavior of `join` is:

```
In [2]: t = Table().with_columns('x', [1, 2, 3], 'y', [10, 11, 12])
In [13]: t2 = Table().with_columns('x', [1, 1, 3], 'z', [100, 105, 110])

In [12]: t
Out[12]:
x    | y
1    | 10
2    | 11
3    | 12
In [14]: t2 # t2 has x = 1 twice
Out[14]:
x    | z
1    | 100
1    | 105
3    | 110

# But when we join, only the first 'x' value in t2 is kept
In [15]: t.join('x', t2)
Out[15]:
x    | y    | z
1    | 10   | 100
3    | 12   | 110

# Same for the opposite join
In [17]: t2.join('x', t)
Out[17]:
x    | z    | y
1    | 100  | 10
1    | 105  | 10
3    | 110  | 12
```

This behavior is a bit strange but it tends to work out for students. I think we should clarify this behavior in the docs and probably add examples demonstrating this.

datascience/tables.py
```diff
@@ -1155,12 +1155,54 @@ def stack(self, key, labels=None):
...
+        a    | b    | c    | d    | e
+        1    | 10   | 6    | 2    | 4
+        9    | 1    | 3    | 1    | 3
+        >>> table.join('a', table2, 'a') #Equivalent to previous join
+        a    | b    | c    | d    | e
+        1    | 10   | 6    | 2    | 4
+        9    | 1    | 3    | 1    | 3
+        >>> table.join('a', table2, 'd') #Repeat column labels relabeled
+        a    | b    | c    | a_2  | e
+        1    | 10   | 6    | 9    | 3
```
SamLau95: Yay for examples :D

maxwelljweinstein:
@SamLau95

datascience/tables.py
```diff
@@ -1196,13 +1197,35 @@ def join(self, column_label, other, other_label=None):
...
-        >>> table.join('a', table2, 'a') #Equivalent to previous join
+        >>> table.join('a', table2, 'a') # Equivalent to previous join
         a    | b    | c    | d    | e
         1    | 10   | 6    | 2    | 4
         9    | 1    | 3    | 1    | 3
-        >>> table.join('a', table2, 'd') #Repeat column labels relabeled
+        >>> table.join('a', table2, 'd') # Repeat column labels relabeled
         a    | b    | c    | a_2  | e
         1    | 10   | 6    | 9    | 3
+        >>> table2 #table2 has three rows with a = 3
```
SamLau95: typo: You meant `a = 1`, right?

datascience/tables.py
```diff
@@ -1196,13 +1197,35 @@ def join(self, column_label, other, other_label=None):
...
+        1    | 2    | 4
+        1    | 2    | 5
+        1    | 10   | 6
+        >>> table #table has only one row with a = 1
+        a    | b    | c
+        9    | 1    | 3
+        3    | 2    | 4
+        3    | 2    | 5
+        1    | 10   | 6
+        >>> table2.join('a', table) # When we join, we get all three rows in table2 where a = 3
```
SamLau95: Here too

maxwelljweinstein: oops, thanks for the catch

SamLau95:
Nice :) Merging.

maxwelljweinstein:
cool thanks!

PR #255, opened 2017-01-01
Tables create retain class of self to support inheritance.

deculler:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -615,7 +615,7 @@ class Table(collections.abc.MutableMapping):
         5                | 6
         """
         labels = self._varargs_as_labels(column_label_or_labels)
-        table = Table()
+        table = self.__class__()
         for label in labels:
             self._add_column_and_format(table, label, np.copy(self[label]))
         return table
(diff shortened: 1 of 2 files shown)


deculler:
John,
   Please take a look at this.  I've included tests in the PR.  It opens up paths for Table.

If we take this step, it will be possible both to refactor tables into a Table_base that has only the container and derived Table that includes the existing viz (inheriting Table_base).  It also makes it possible to build upon tables.  I've done an ML_Table that brings in the scikit-learn functionality needed to do everything in "Introduction to Statistical Learning with applications in R".  You can see that (in progess) at https://github.com/deculler/MachineLearningTables.

deculler:
Not wanting to nag, but please let me know what I can do to move this along.  Being a container that can't be inherited from really limits its utility.  It is particularly problematic with our (correct) view that most operations return a new object, rather than modify it.  The problem is that new object is a Table, rather than the subclass.  Thus, for any subclass object X, type(X) != type(X.copy()) - or select, take, where, ... for all the other non-mutating methods.  The test code shows problems with the current implementation.

deculler:
Great

On Feb 3, 2017 4:33 PM, "John DeNero" <notifications@github.com> wrote:

*@papajohn* commented on this pull request.

Great! I just need to merge & change self.__class__ to type(self) to be
conventional, then I'll merge. Sorry it took a little while.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/data-8/datascience/pull/255#pullrequestreview-20132507>,
or mute the thread

papajohn:
Pushed the change: https://github.com/data-8/datascience/commit/780eb365634c1715daaf80110d80686c62ce3eb5

PR #258, opened 2017-01-13
Fixes bug in Issue 204

adnanhemani:
Check arguments for Table.append to only accept row with the same amount of elements as columns of the table

[*] Wrote test for feature

**Changes proposed:**
Check the length of the inserting row and make sure it matches the number of columns the table has before adding the row into the table.


adnanhemani:
Made the requested change!

PR #260, opened 2017-01-14
Faster Table.split

adnanhemani:
**Changes proposed:**
Changed Table.split so that it utilizes Table.take, which helps us split the table much faster than copying the rows multiple times.


adnanhemani:
It seems that Coveralls isn't letting me pass because I made the code more concise (and therefore there's less lines to hit). Thoughts?

PR #277, opened 2017-03-01
Fix minor formatting issues with Table.as_html 

Sumukh:
- [ ] Wrote test for feature 
- [ ] Added changes in the Changelog section in README.md
- [ ] Bumped version number (delete if unneeded)

I did not do any of the above - sorry. Writing a test would be a good idea ... maybe I'll add one. 

I came across these when working on table rendering for okpy.org (https://github.com/Cal-CS-61A-Staff/ok/pull/1099)

**Changes proposed:**
- 92ed686 fixes the `</p>` tag that occurs when rows in the table are omitted
- 38dcdc8 makes the `</tbody>` tag appear only once per table. Currently one appears for every row - despite only having one opening `<tbody>` tag.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1872,8 +1872,8 @@ class Table(collections.abc.MutableMapping):
                 (3, ' '.join('<td>' + fmt(v, label=False) + '</td>' for
                     v, fmt in zip(row, fmts))),
                 (2, '</tr>'),
-                (1, '</tbody>'),
             ]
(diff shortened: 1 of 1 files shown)


davidwagner:
It looks like these fixes have already been merged in commits 1ef10ebde4f8bb2475aba4c49e29ee9aeed8bad6 and 5eaeb4e5c10cb62993e93c8aef7fd0ef6569078e.

PR #278, opened 2017-03-03
Add plt.show() to all plotting calls

SamLau95:
This doesn't change current behavior at all. However, without it the
plot gets displayed multiple times when using IPython's widgets as I'm
discovering now.

I've tested this by calling each plot method and verifying that the
notebook output is the same. In addition, I've verified that using IPython
widgets does not duplicate plots when redrawn:

interact


SamLau95:
Nevermind. I realized this does break existing notebooks since you can't modify the plot (eg. with `plt.title()`) after you call `plt.show()`. Instead, I can call `plt.show()` manually in functions I want to make interactive.

Issue #286, opened 2017-05-05
Add move_column(label, index) that moves a column but leaves the rest alone


papajohn:

papajohn:
FYI @a-adhikari

adnanhemani:
Closed by #291

PR #291, opened 2017-05-16
move_column implemented

adnanhemani:
[* ] Wrote test for feature

**Changes proposed:**
As per issue #286.


README.md
```diff
@@ -27,6 +27,9 @@ pip install datascience
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.9.6
+- 'move_column' added.
+
```
davidwagner: We'll need to fix the version number.

datascience/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.9.5'
+__version__ = '0.9.6'
```
davidwagner: We'll need to fix the version number.

datascience/tables.py
```diff
@@ -1795,6 +1795,16 @@ def bin(self, *columns, **vargs):
             binned[label + ' ' + tag] = np.append(counts, 0)
         return binned
 
+    def move_column(self, label, index):
+        """Returns a new table with specified column moved to the specified column index."""
+        table = type(self)()
+        col_order = list(self._columns)
+        label = col_order.index(self._as_label(label))
+        col_order.insert(index, col_order.pop(label))
```
davidwagner: This has two side-effecting operations on `col_order` in the same statement, which seems unnecessarily tricky.  I suggest rewriting as `col = col_order.pop(label)` then `col_order.insert(index, col)`.  Also I'd find this clearer as `label_idx = col_order.index(self._as_label(label))` and using `label_idx` later, rather than overwriting `label` with a value of a different type.

Issue #287, opened 2017-05-08
Should predicates throw warnings on arrays?


papajohn:
Students often write `t.where(1, are.equal_to(t.column(2))`, which doesn't work. There's no good way to get it to work. The correct syntax is `t.where(1, are.equal_to, 2)`. Perhaps a warning or error should redirect them.

adnanhemani:
Closed by #292. #370 will stay open for implementation clarification in the meantime.

PR #292, opened 2017-05-16
Warning added for incorrect Predicate use

adnanhemani:
[* ] Wrote test for feature

**Changes proposed:**
As described in issue #287.


datascience/predicates.py
```diff
@@ -237,5 +248,13 @@ def _equal_or_float_equal(x, y):
...
         return x == y
 
+############
+# Utility #
+############
+def check_iterable(y):
+    if hasattr(y, '__iter__'):
+            warnings.warn("You are checking against an iterable. If you are trying to test one iterable against another \
+                using 'Table.where', put your first iterable as your first argument, your predicate as your second  \
+                argument, and then your second iterable as your third argument.")
```
davidwagner: Let's see if we can improve the warning.  I'm not sure if this will mean anything to students ("iterable", "checking against an iterable".)  How about "Do not pass an array or list to a predicate.  If you are trying to find rows where two columns are the same, use table.where('c', are.equal_to, 'd') instead of table.where('c', are.equal_to(table.column('d')))."

README.md
```diff
@@ -27,6 +27,10 @@ pip install datascience
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.9.6
+
+- Warning added for comparing iterables using predicates incorrectly.
+
```
davidwagner: We'll need to fix the version number before merging.

datascience/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.9.5'
+__version__ = '0.9.6'
```
davidwagner: We'll need to fix the version number before merging.

tests/test_tables.py
```diff
@@ -266,6 +266,16 @@ def test_where_predicates(t):
...
+def test_where_predicates_warning(t, capsys):
+    t1 = t.copy()
+    t1['count1'] = t1['count'] - 1
+    with (pytest.raises(ValueError)):
+        test = t1.where('count', are.equal_to(t1.column("count1")))
+    out, err = capsys.readouterr()
+    assert "You are checking against an iterable. If you are trying to test one iterable against another \
+                using 'Table.where', put your first iterable as your first argument, your predicate as your second  \
+                argument, and then your second iterable as your third argument." in err
+
```
davidwagner: Update the assert statement to match the updated warning.

datascience/predicates.py
adnanhemani: Agreed! I've made the change

Issue #280, opened 2017-04-21
Group by category for hist


papajohn:
It's a pain to draw two overlaid histograms that differ by a categorical variable. E.g., to overlay these two:

```
nonsmokers = baby.where('Maternal Smoker', are.equal_to(False))
baby.where('Maternal Smoker', False).hist('Birth Weight')
baby.where('Maternal Smoker', True).hist('Birth Weight')
```

the simplest expression is

```
baby.where('Maternal Smoker', False).bin('Birth Weight')\
  .join('bin', baby.where('Maternal Smoker', True).bin('Birth Weight')).\
  relabeled(1, 'False').relabeled(2, 'True').hist(bin_column='bin')
```

papajohn:
Proposed syntax

`baby.hist('Birth Weight', group='Maternal Smoker')`

a-adhikari:
Great! Agree that it's needed. Could also be used to overlay histograms using data from two different tables, e.g. distribution of incomes from Table 1 and distribution of incomes from Table 2.
Ani

--
Ani Adhikari
Teaching Professor
Department of Statistics
University of California, Berkeley

henryem:
@papajohn I'd like to use this, and I'd be happy to implement it.

papajohn:
Great!

henryem:
Actually @papajohn, I happened to find `pivot_hist` just now.  Is this the point of `pivot_hist`?  The issue is that `pivot_hist` really makes a bar graph, with each bin getting side-by-side bars for each category, instead of overlaying the histograms.  (I guess that's why it's called `pivot_hist` rather than `group_hist`, since it thinks of the bins as categories.)

It's confusing to have both, and I wonder if the behavior of `pivot_hist` really makes sense.  Do you know if it's used?  We could have an option to recover the `pivot_hist` behavior, like `t.hist('bar', bins='foo', side_by_side=True)`.

papajohn:
As far as I know, pivot_hist is not used. I agreed that it's an unusual
thing and doesn't fit nicely into the rest of the module.

henryem:
Closed by #293.

Issue #242, opened 2016-09-27
Bar chart from categories


papajohn:
From @a-adhikari : `barh` should surely take a column of categorical data and draw the bar graph, without us having to do the grouping for it first. That would be consistent with hist having a default behavior when we don't specify bins.

papajohn:
We should resolve #233 first.

papajohn:
I think we'll need a new method for this behavior, since `barh` on one argument already has a useful behavior that's consistent with `scatter` and `plot`. 

Matlab calls the function that displays a categorical distribution a "histogram"

https://www.mathworks.com/help/matlab/matlab_prog/plot-categorical-data.html?requestedDomain=www.mathworks.com

henryem:
@papajohn I'd like to have this for Data 8R this summer, and I'd be happy to implement it.

Is there a reason to resolve #233 first?  Presumably we'll need to preserve the public API for the many existing plotting methods anyway, and this is a common enough use case that I think it's worth supporting.

I think it's too confusing to call this a histogram, but I can't think of a clever name.  Perhaps `group_barh` (and `group_bar`)?  Another option is to make this the behavior of `t.barh(category_column='foo')`.  If either of those seems okay, I'll implement it.

papajohn:
You can add it without #233 . `group_bar` is fine.

henryem:
Closed by #293 in 0.10.0.

PR #293, opened 2017-05-20
Grouped charts

henryem:
[x] Wrote test for feature (in Charts.ipynb as with other viz code)

**Changes proposed:**
Addresses #280 and #242.  Also refactors the `hist` code to be (IMO) slightly cleaner.


henryem:
@SamLau95 these (non-breaking) changes may be relevant for this summer's Data 8.

The addition of `group_barh` will simplify some code slightly and allow you to talk about `barh` before `group`, if you want.

The addition of grouping to `hist` might open up some cross-categorization examples that were previously too cumbersome to use.  Note that it produces the conditional distribution of each group, not the marginal; that means that each group's histogram sums to 1, rather than to the proportion of rows in the original table that are in the group.

henryem:
To be clear, the coverage test fails because there aren't (as far as I can tell) automated tests for visualizations, but I've added to the manually-runnable tests in Charts.ipynb.

papajohn:
@samlau95 can you review, since you and Henry are the next users of the
module?

datascience/tables.py
```diff
@@ -2052,6 +2052,29 @@ def annotate(axis, ticks):
 
         self._visualize(column_for_categories, labels, xticks, overlay, draw, annotate, width=width, height=height)
 
+
+    def group_bar(self, column_label, **vargs):
+        """Plot a bar chart for the table.
+
+        The values of the specified column are grouped and counted, and one
+        bar is produced for each group.
```
SamLau95: Let's leave a note here that although this method is just a `group` followed by a `bar`, we use it for pedagogical purposes. That way it's a bit more clear why this method exists.

datascience/tables.py
```diff
@@ -2132,6 +2155,29 @@ def annotate(axis, ticks):
 
         self._visualize('', labels, yticks, overlay, draw, annotate, width=width, height=height)
 
+
+    def group_barh(self, column_label, **vargs):
+        """Plot a horizontal bar chart for the table.
+
+        The values of the specified column are grouped and counted, and one
+        bar is produced for each group.
```
SamLau95: Let's leave a note here that although this method is just a `group` followed by a `barh`, we use it for pedagogical purposes. That way it's a bit more clear why this method exists.

datascience/tables.py
henryem: Okay, I've added a comment in my next commit.  I think this method is actually a good addition to the API, not just for pedagogical purposes.  Its behavior is more intuitive when you're thinking about `bar` as displaying a distribution (like `hist`) rather than a series of points with heights (like `scatter`).  But I agree it will cause confusion to have both methods.  Hopefully the comment will minimize that.

datascience/tables.py
henryem: (See above.)

henryem:
Thanks, @SamLau95.

PR #301, opened 2017-11-21
Updating requirements.txt with all dependencies

jasonyzhang:
[ ] Wrote test for feature
[ x ] Added changes in the Changelog section in README.md

**Changes proposed:**
Added all dependencies for `datascience` so that `pip install -r requirements.txt` works for new envs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ pip install datascience
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.10.4
+
+- Updated `requirements.txt`
+
 ### v0.10.3
 
 - Fix `OrderedDict` bug in `Table.hist`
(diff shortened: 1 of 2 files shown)


papajohn:
For now, `datascience` is a module used primarily by newcomers, and due to `pip` defaults, fully declaring the dependencies of `datascience` has caused newcomers to `pip` various problems. E.g., attempting to upgrade `datascience` often leads one to accidentally upgrade `matplotlib` and `scipy`. For this reason, we leave out the dependencies that are part of a standard Anaconda build.

I do agree that the dependencies should be listed somewhere, though perhaps not in the `requirements.txt` file.

davidwagner:
Closing as won't-merge per papjohn's comments above.

Issue #305, opened 2017-12-13
Implement join on multiple columns


papajohn:

adnanhemani:
Implemented by #306. @TomPiona (Ronnie), a job well done! :)

PR #306, opened 2017-12-16
Joining on multiple columns

TomPiona:
[x] Wrote test for feature
[x] Bumped version number (please change if incorrect)

**Changes proposed:**
Addressing #305.  In addition to the existing way of joining, now a list/array of columns can be passed in to join on.

A question on behavior and one on design:

- Behavior: Right now, when a label in the second table does not match the label in the first, I have it drop the second column label and use the first.  Pandas keeps both columns if the labels don't align, so I was wondering if there was a preference.
- Design: I split `join` into two functions, moving most of the original function to `_join` and the new stuff to `_multiple_join`, but there's a lot of repetitive code. Would it be better to leave everything under join and use some more control statements?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ pip install datascience
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.11.0
+
+- Added `join` for multiple columns.
+
 ### v0.10.3
 
 - Fix `OrderedDict` bug in `Table.hist`
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1179,13 +1179,13 @@ class Table(collections.abc.MutableMapping):
         rows for all values of a column that appear in both tables.
 
         Args:
-            ``column_label`` (``str``):  label of column in self that is used to
+            ``column_label``:  label of column or array of labels in self that is used to
                 join  rows of ``other``.
             ``other``: Table object to join with self on matching values of
                 ``column_label``.
 
         Kwargs:
-            ``other_label`` (``str``): default None, assumes ``column_label``.
+            ``other_label``: default None, assumes ``column_label``.
                 Otherwise in ``other`` used to join rows.
 
         Returns:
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -27,6 +27,10 @@ pip install datascience
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.11.0
+
+- Added `join` for multiple columns.
+
```
davidwagner: We'll need to update this before merging.

datascience/tables.py
```diff
@@ -1238,12 +1238,31 @@ def join(self, column_label, other, other_label=None):
...
+            # then we are going to be joining multiple labels
+            return self._multiple_join(column_label, other, other_label)
+
+        # original single column join
+        return self._join(column_label, other, other_label)
+
+    def _join(self, column_label, other, other_label=None):
+        """joins when COLUMN_LABEL is a string"""
+        if self.num_rows == 0 or other.num_rows == 0:
+            return None
```
davidwagner: What's the purpose of the check in lines 1261-1262?  Do we need this?  I would have expected that if you try to join an empty table to some other table, it would be more intuitive to return an empty table rather than None.  Is that right?  I'm not sure we should be changing the behavior of .join() on an empty table -- that seems like a change that could potentially break prior code.

datascience/tables.py
```diff
@@ -1273,6 +1292,43 @@ def join(self, column_label, other, other_label=None):
...
+        self_rows = self._multi_index(column_label)
+        other_rows = other._multi_index(other_label)
+
+        # Gather joined rows from self_rows that have join values in other_rows
+        joined_rows = []
+        for v, rows in self_rows.items():
+            if v in other_rows:
+                joined_rows += [row + o for row in rows for o in other_rows[v]]
+        if not joined_rows:
+            return None
```
davidwagner: Should this be returning an empty table, rather than None?

davidwagner:
This looks very nice.  I'm a little wary of the fact that so much code has been duplicated (copy-pasted) in both `_join()` and `_multiple_join()`.  That looks dangerous from a maintenance point of view (I'm worried we'll modify one of them without remembering to make the same change to the other).  Can we refactor this to avoid repeated code (DRY principle); or to make `_join(self, column_label, other, other_label)` check that `other_label` is a string and then  call `self._multiple_join([column_label], other, [other_label])`?

PR #309, opened 2018-01-10
fixed numpy arrays to be appended

adnanhemani:
[*] Wrote test for feature
[*] Added changes in the Changelog section in README.md
[*] Bumped version number (delete if unneeded)

**Allow NumPy arrays to be used for append:**
Up to now, NumPy arrays would be caught and not allowed to be used with Table.append. With this PR, this would now be allowed.


README.md
```diff
@@ -27,6 +27,10 @@ pip install datascience
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.10.4
+
+- Allow NumPy arrays to be appended into tables.
+
```
davidwagner: We'll need to fix the version number.

PR #313, opened 2018-04-17
Update tables.py

yufengg:
unused variable; looks like the intention was to have it be used in the for loop for efficiency? If not, looks like the placeholder var is not needed.

Tiny implementation tweak, not sure we need to include in changelog nor bump version number... please advise.

[ ] Wrote test for feature

**Changes proposed:**
Included for-loop placeholder variable usage.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -151,7 +151,7 @@ class Table(collections.abc.MutableMapping):
         """Convert a Pandas DataFrame into a Table."""
         t = cls()
         labels = df.columns
-        for label in df.columns:
+        for label in labels:
             t.append_column(label, df[label])
         return t
 


datascience/tables.py
```diff
@@ -151,7 +151,7 @@ def from_df(cls, df):
         """Convert a Pandas DataFrame into a Table."""
         t = cls()
         labels = df.columns
-        for label in df.columns:
+        for label in labels:
```
davidwagner: I think I'd prefer a change that insteaddeletes the unused `labels = df.columns` line.  I don't see why this would make any difference in efficiency.

adnanhemani: I think it was left this way for readability - "df.columns" isn't as descriptive as "labels" for what it really is. Thoughts?

davidwagner: OK, seems fine to me.

Issue #324, opened 2018-09-10
Matplotlib upgrades break formatting and cause warnings


rameshvs:
* Histogram bars no longer have borders ([default style change](https://matplotlib.org/users/dflt_style_changes.html#patch-edges-and-color))
* Bar chart labels are now centered by default, so the workarounds in `Table.bar` and `Table.barh` now mess things up
* `plt.hist(normed=...)` has been deprecated in favor of `plt.hist(density=...)`, which causes a scary pink warning box to pop up every time a histogram is drawn.

PR #325, opened 2018-09-10
Address formatting problems from matplotlib 2.x upgrade

rameshvs:
[ ] Wrote test for feature (*Couldn't due to lack of GUI tests*)

## Changes proposed:
Addresses some minor formatting issues and warnings introduced by upgrading to matplotlib 2.x. See https://github.com/data-8/datascience/issues/324 for details.

## To test:

1. Generate a histogram. Before (i.e., on `f73aaf45f8b3`), you'll see a warning about the `normed` keyword and no lines around the bars; both these issues should be fixed after.

2. Generate a bar chart. Before (i.e., on `f73aaf45f8b3`), you'll see the labels are offset from the bars, but this issue should be fixed after.


rameshvs:
I had one test fail locally (`test_scatter_error`), but was unable to reproduce the failure either at the command line or in a notebook.

Issue #318, opened 2018-06-06
Create a new syntax for histograms with counts on the y axis


papajohn:
Right now, `t.hist(x, normed=False)` is the syntax for generating a histogram with counts on the vertical axis. That's confusing... what does `normed` mean? How about:

`t.hist_of_counts(x)`

davidwagner:
See PR #326

adnanhemani:
Closed as per #327.

PR #327, opened 2018-09-12
Add counts_hist function

rameshvs:
[x] Wrote test for feature

**Changes proposed:**
Add `hist_of_counts` function, as in https://github.com/data-8/datascience/issues/318.


rameshvs:
@papajohn @davidwagner I can't seem to request a review or assign anyone, but here's the PR for #318 .

datascience/tables.py
```diff
@@ -2569,6 +2569,102 @@ def draw_hist(values_dict):
...
+        >>> t = Table().with_columns(
+        ...     'count',  make_array(9, 3, 3, 1),
+        ...     'points', make_array(1, 2, 2, 10))
+        >>> t
+        count | points
+        9     | 1
+        3     | 2
+        3     | 2
+        1     | 10
+        >>> t.hist() # doctest: +SKIP
```
papajohn: Is this meant to be `t.hist_of_counts()`, or is it just stale and should be deleted?

datascience/tables.py
```diff
@@ -2569,6 +2569,102 @@ def draw_hist(values_dict):
...
+        >>> t = Table().with_columns(
+        ...     'value',    make_array(1,   2,   3,   2,   5  ),
+        ...     'category', make_array('a', 'a', 'a', 'b', 'b'))
+        >>> t.hist('value', group='category') # doctest: +SKIP
+        <two overlaid histograms of the data [1, 2, 3] and [2, 5]>
+        """
+
+        # TODO test case bins = np.arange(3, 40, 5/3)
+        # TODO test case bins = np.arange(1e-14, 6e-14, 1e-13)
+        # TODO test case bins = np.arange(1e-20, 2e-20, 1e-21)
```
papajohn: Looks like some of these are done already.

rameshvs:
Thanks for the catch! Comments addressed. I also changed one of the docstring examples to use counts, and changed it so that when using `bin_column`, it checks whether the counts are integers (or close to integers). 

The check only works if columns are passed in: fixing it would probably require a bit of refactoring to avoid duplication between `hist` and `hist_of_counts`.

Issue #323, opened 2018-09-10
Tests failing because default formatter changed


davidwagner:
Some of the tests with `make test` are failing, because the default format for numbers changed: instead of displaying numbers as `12345`, the default formatter now displays them as `12,345`.  Is there any reason why this is a problem?  If not, we should just update the tests.

papajohn:
I think we should just update the tests.

davidwagner:
OK.  See PR #331: it should fix this.

davidwagner:
Gaah! It looks like I've just caused problems here.  I tested on my local system but not on Travis, and the two don't behave identically.  Now that PR #331 has been merged, the tests pass on my local build, but they fail on Travis (e.g., https://travis-ci.org/data-8/datascience/builds/430716572).

Looking into this further, it seems to be a difference between Python 3.6 vs 3.7.  With Python 3.6, 123456 gets formatted as `123,456`.  With Python 3.7, it gets formatted as `123456`.  My local system has 3.6 installed.  Travis seems to be using 3.7.  (On the surface `.travis.yml` seems to specify 3.6, but I think using `Miniconda3-latest` overrides that with the latest version of Python, which is now 3.7.)

I apologize.  Rather than fixing a problem, I seem to have created a problem.  How would you like to resolve this?  Should I generate a pull request that reverses the test cases to use the format that will appear on Travis and Python 3.7 (back to where we were before merging PR #331)?

deculler:
We really don't want Tables to be specific to.one version of python.  We
had better fix the test to accept either.

davidwagner:
OK, I'm very confused.  It isn't Python 3.6 vs 3.7 that explains the difference after all.  Still trying to narrow down what is the difference between these two Python environments.  (Current hypothesis: it might be locale-specific.)

deculler:
Yes, they behave differently but we want Tables to be available and pass
its regression tests in either.  We do need to limit what versions we
support, but it would seem we need to accommodate both in the tests.  It is
used far beyond our offering of a course.

adnanhemani:
As is, this library will not even run the tests due to errors in formats.py in Python 3.4. Unsure of how to fix this, but I have confirmed it is an issue with backwards compatibility in py3.4 - when I upgraded my virtual environment to py3.7, this was no longer an issue. Pasting my error code below for future reference:

___________________________________________________ ERROR collecting tests/test_formats.py ____________________________________________________
[...]

PR #329, opened 2018-09-12
Update dependency version numbers

### LINKED ISSUE #328: Update dependencies to specify newer version of setuptools
I think we need to update the version of the `setuptools` package used.  `linux_environment.yml` specifies setuptools 19.6.2, and I think that's causing the Travis build to break.

The Travis build goes awry in running `python3 setup.py install`; it tries to `pip install folium==0.2.1`, which triggers an install of `requests>=1.0.0`, which triggers an install of `docopt>=0.6.1`, which triggers an install of `PyYAML>=3.10`, which triggers an install of `pathlib2>=2.2.0`, which triggers the following error message:

``error: Setup script exited with error in pathlib2 setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected version spec in scandir;python_version<"3.5" at ;python_version<"3.5"``

It appears that this is an [environment marker](https://www.python.org/dev/peps/pep-0508/#environment-markers), and I think setuptools 19.6.2 doesn't support environment markers.

I think we need to update the dependencies to use a newer version of setuptools.  setuptools 19.6.2 is a pretty ancient version of setuptools; the current is 40.2.0.  From the [setuptools version logs](https://setuptools.readthedocs.io/en/latest/history.html), it looks like support for environment markers was only added in setuptools 20.5.

In general, a lot of the dependencies in `linux_environment.yml`, `osx_environment.yml`, `readthedocs.yml` are very old (Python 3.4, etc.).  We should probably just update them all.  Also `requirements.txt` specifies folium 0.2.1 and we should probably update that to avoid specifying any version number so we get the latest version of folium.

davidwagner:
**Changes proposed:**
I've updated the dependencies in `linux_environment.yml` and `osx_environment.yml`. Also updated to Python 3.6 in `.travis.yml` and `readthedocs.yml`.  Hopefully this will fix #328, so it builds with Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
 sudo: false
 
 python:
-  - "3.4.2"
+  - "3.6"
 
 cache:
   apt: true
--- a/linux_environment.yml
+++ b/linux_environment.yml
@@ -1,51 +1,51 @@
 name: datascience
 dependencies:
-- cycler=0.9.0=py34_0
-- decorator=4.0.6=py34_0
-- fontconfig=2.11.1=5
-- freetype=2.5.5=0
-- ipython=4.1.1=py34_0
-- ipython_genutils=0.1.0=py34_0
-- libgfortran=1.0=0
-- libpng=1.6.17=0
-- libxml2=2.9.2=0
-- matplotlib=1.5.1=np110py34_0
-- mkl=11.3.1=0
-- numpy=1.10.4=py34_0
-- openssl=1.0.2f=0
-- pandas=0.17.1=np110py34_0
-- path.py=8.1.2=py34_1
-- pexpect=3.3=py34_0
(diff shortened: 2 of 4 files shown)


davidwagner:
Looks like this does indeed un-break Travis (so now Travis can build again).  It does expose a separate bug (#323).

davidwagner:
BTW, I didn't update the folium dependency; I left that hardcoded at version 0.2.1, because I don't know what the impact of that change might be.

papajohn:
I'm blindly approving this one because I really don't know the consequences, but glad to hear that Travis will work again.

SamLau95:
This seems safe to me - I haven't heard of any breaking updates for the packages we depend on. Having Travis work should give us more peace of mind.

PR #330, opened 2018-09-12
Fix test_scatter_error. 

davidwagner:
**Changes proposed:**
test_scatter_error() looks like a bogus test to me.  I'm not sure how this test ever worked.  One of the example tables *does* have a column called `letter`, so it shouldn't raise a ValueError.  Correct the test so it will fail properly.

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1261,7 +1261,7 @@ def test_scatter_error(table):
     non-numerical values."""
 
     with pytest.raises(ValueError):
-        table.scatter('letter')
+        table.scatter('nonexistentlabel')
 
 def test_df_roundtrip(table):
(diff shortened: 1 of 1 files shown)


rameshvs:
I thought `scatter` is supposed to raise a `ValueError` because the `letter` column is non-numeric (based on "ValueError -- Every column, ``column_for_x`` or ``select``, must be numerical." in the docstring). 

That being said, it doesn't actually seem to do that, and I don't see any additional checks inside of `scatter`. `_visualize` only checks the y columns, so this seems to slip through the cracks.

papajohn:
Once upon a time, I think the scatter method probably checked for numeric values. Disabling it in this way seems fine for now.

PR #336, opened 2018-10-03
Add dependencies to requirements.txt

### LINKED ISSUE #335: Datascience doesn't declare its dependencies
I think that Datascience needs the following packages to import correctly:

* matplotlib
* pandas
* scipy
* ipython

These aren't declared in `setup.py` or similar, I think.  At least, they do not get installed with `pip install datascience`.  Worth adding?

davidwagner:
[ ] Wrote test for feature

**Changes proposed:**
Fixes #335 by adding dependencies

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,7 @@
 folium==0.2.1
 sphinx
 setuptools
+matplotlib
+pandas
+scipy
+ipython


ryanlovett:
Would it be possible to annotate the requirements with specific (==), minimum (>), and/or maximum (<) version strings? Sometimes other packages will want to pull in newer versions of these dependencies and it'd help to know whether datascience is compatible.

rameshvs:
I think I saw a previous PR / comment that discouraged adding these to `requirements.txt` since the recommended installation method is through anaconda (which provides all these packages anyway).

davidwagner:
@rameshvs , do you happen to remember where you saw that?  I have the impression that the datascience library is used outside of Data 8, so I'm not sure whether we want to assume it will be installed through conda.  That said, this is not something I feel strongly about it.

deculler:
It is extremely important that it be straight forward to use outside data8,
because otherwise it becomes both a stumbling block to adoption.of the
pedagogy and something that has to be defended.  "Why are you teaching
Pandas?".

davidwagner:
John suggests we should not list all dependencies in requirements.txt: see https://github.com/data-8/datascience/pull/301#issuecomment-380928606.

Issue #314, opened 2018-04-23
append_column does not return the table


papajohn:
The `append_column` method should return the table for method chaining, but it doesn't.

adnanhemani:
Closed as per #344

PR #344, opened 2019-02-01
Fixed append_columns() function

nnj1:
This function, according to the docs, modifies the table in place but also returns a handle to the table for further chaining. This behavior did not work as expected because someone probably forgot to return self.

[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -489,6 +489,7 @@ class Table(collections.abc.MutableMapping):
             self._num_rows = len(values)
 
         self._columns[label] = values
+        return self
 
     def relabel(self, column_label, new_label):
         """Changes the label(s) of column(s) specified by ``column_label`` to


davidwagner:
I suspect you mean that it fixes append_column (not append_columns).  This should fix #314.

davidwagner:
Check the examples in the docstring -- I suspect the output will now be different from what's listed in the docstring for append_column (namely, `table.append_column(...)` will now show the modified table after that cell; and then there will be no need to view the modified table in the next cell using `table`).  We should update the docstring to match the actual behavior of the code.

davidwagner:
Thanks much for your submission!  We've made changes based on your pull request (see PR #358).  I appreciate your help with this library.

PR #345, opened 2019-02-23
fixed bug

adnanhemani:
[* ] Wrote test for feature

**Changes proposed:**
There is a bug where if you join tables together where if the second table already has a duplicated column (perhaps from a previous join), you will get the wrong values (and also not enough columns) in your join. A Jupyter notebook with corresponding CSVs is attached showing this problem - notice how there are fewer columns in the first join (and the values of that column are wholly incorrect).

I've made a quick workaround making it so that we generate new labels if such a problem is detected while joining.


adnanhemani:
@papajohn maybe you'd be a good reviewer for this?

adnanhemani:
Sorry, here are the files mentioned.
[broken_join.tar.gz](https://github.com/data-8/datascience/files/2896443/broken_join.tar.gz)

davidwagner:
Alas, this went over my head.  The example in the notebook was too complicated for me to follow, and I don't feel confident in my ability to understand the changed code.  (I'm not even familiar with the basics of how duplicated columns can arise, let alone what should happen when joining a table containing them or what the semantics ought to be.)   So, I can't provide any useful review.  Perhaps we can find someone else to review this.

davidwagner:
Is it possible to add a simple test case that takes two tables (the simplest possible to trigger this behavior), joins them, and asserts what the result should be?  (By converting the table to a string and asserting equality to the desired output.)

adnanhemani:
Here's a simpler Jupyter notebook with the issue demonstrated. I'll be making minor changes to the PR code as per this.
[broken_join_simplified.ipynb.zip](https://github.com/data-8/datascience/files/3292966/broken_join_simplified.ipynb.zip)

davidwagner:
Cool, that helped me understand, and now this is looking good.

Might you be willing to write a second function with a test case that executes In[2]-In[4] and then checks the result, in addition to the test you already have?  In other words, it creates tables `scrabble` and `scrabble_2` and then executes `scrabble.join("letter", scrabble_2)` and uses `assert_equal()` to check that the result is equal to the correct value for that table?  That would give us a separate, very simple test case that is violated by the current semantics but is satisfied by the corrected code, might make it easier to figure out what's going wrong if we ever have a regression, and documents the intended behavior cleanly.  The current `test_join_with_column_already_duplicated` function is a more powerful test, but it's also harder to understand because there are a lot of moving parts.

adnanhemani:
I've added the test, I agree with the suggestion! Let me know if this is what you were imagining.

davidwagner:
Perfect!  Thank you!

Issue #352, opened 2019-06-03
pandas upgrade causes read_table warnings


ktakimoto:
<code>read_table</code> now causes pink warning boxes in notebooks since <code>pd.read_table</code> has been deprecated in favor of <code>pd.read_csv</code>

adnanhemani:
I've done some reading on the Pandas GH, it seems like the read_table method will now be getting [un-deprecated in v0.25.0](https://github.com/pandas-dev/pandas/issues/25220) (release scheduled for 07/01/19). As such, I'd recommend to not make any changes and simply upgrade to the latest version of Pandas when it is released. AFAIK, datahub is running Pandas v0.23.4, so the deprecation warning should not show up there at all.

davidwagner:
Have the pandas folks documented that decision anywhere?  I don't see any decision in that link to the issue over on Pandas to undeprecate read_table.

deculler:
Perhaps we should weigh in on the side of maintain read_table with its
broader notion than csv to help move this to close. ( If read_csv does more
than what it says, so be it.)  A rather discouraging dialog and no clear
conclusion.  Is there a way to suppress the warning while they go back and
forth on this?

adnanhemani:
@davidwagner Yes, if you go to the bottom of the issue that I linked to above, the Pandas folks have added the issue to the v0.25.0 release and the last comment says something similar.

@deculler as far as I know (and based off the testing I’ve done) there is no warning in Datahub at this time (since the version of pandas installed on that server is a couple versions old). Are you seeing this on there or locally?

deculler:
There was datahub warnings during the pedagogy bootcamp last week, but may
been resolved by rollback.  Our dependence is so light that we can
certainly wait this one out and see what they do.

davidwagner:
I'm inclined to think we should change read_table to read_csv in the datascience library.  I don't see any downside, as it seems to have the same functionality and is just a different name for the code; and it makes datascience work more cleanly with all versions of pandas, so it takes this issue off the table.

adnanhemani:
I'm mainly scared of an API change on this function since it's so vital to this library and I don't understand the edge cases well enough - I have a feeling any change will affect the edge cases (non csv files that are currently supported). But I can work on a patch if everyone else thinks it might be best to switch over.

deculler:
I think it is unlikely that pandas folks will remove non-csv functionality
from read_csv.  Too much already relies on that.  Certainly they won't
[...]

PR #353, opened 2019-06-07
Update read_table to pandas 0.42 read_csv

deculler:
[ X] Wrote test for feature

**Changes proposed:**
Updates read_table function to use pandas.read_csv rather than pandas.read_table, which was deprecated 0.24

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -129,7 +129,7 @@ class Table(collections.abc.MutableMapping):
                 vargs['sep'] = ','
         except AttributeError:
             pass
-        df = pandas.read_table(filepath_or_buffer, *args, **vargs)
+        df = pandas.read_csv(filepath_or_buffer, *args, **vargs)
         return cls.from_df(df)
 
     def _with_columns(self, columns):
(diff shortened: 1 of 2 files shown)


adnanhemani:
As I listed in #352 in more detail, I think it would be okay to not make this change (I would actually recommend against for readability purposes only) - it shouldn't affect our deployment on datahub and this deprecation looks like it'll be reversed in the next version of Pandas. If we still want to make this change, I think it should still be okay, though.

adnanhemani:
Related to this, perhaps, we should lock down a fixed Pandas version number in our requirements.txt? I'm noticing that Pandas isn't even listed as a dependency that we rely on in our requirements.txt file - but as our library relies on it to import tables, we probably should have it listed.

davidwagner:
@adnanhemani,  John says he doesn't want to list list pandas in the requirements.txt for datascience, as it causes surprising results when newcomers install datascience via pip.  We could list a version in the datahub config but maybe it's not needed.

deculler:
The CVS reader and to/from dataframes is the full extent of the pandas
dependence.  This has protected us a lot from the many changes there.  We
could probably afford to special case the version in the code till it is
reversed, since it is only the one point. The warning is really distracting
to students and looks bad.  The other question is whether the python csv
reader has caught up.  Each used to fail on various different valid csv
files, but pandas was a little more robust.

davidwagner:
I've merged #362, which is based on this, so this is now part of the library.  Thanks for the changes!

PR #354, opened 2019-06-08
Cleanuptests

deculler:
[ X] Wrote test for feature

**Changes proposed:**
Updates read_table function to use pandas.read_csv rather than pandas.read_table, which was deprecated 0.42

Update default format to include commas in integer formatting, independent of localization.

Cleaned up tests.  All pass.

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1332,3 +1332,7 @@ def test_url_parse():
     with pytest.raises(ValueError):
         url = 'https://data8.berkeley.edu/something/something/dark/side'
         Table.read_table(url)
+
+def test_read_table():
+    """Test that Tables reads a csv file."""
+    assert isinstanceof(Table, Table().read_table("tests/us-unemployment.csv"))
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -94,10 +94,9 @@ def test_doctests():
 # Overview #
 ############
 
-
-def test_basic(t):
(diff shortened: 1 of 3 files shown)


tests/test_tables.py
```diff
@@ -267,8 +284,13 @@ def test_where_predicates(t):
     """)
 
 
-def test_sort(t):
+def test_sort(table):
+    t = table
+    t['totals'] = t['points'] * t['count']
+    print("test_sort")
```
adnanhemani: We should probably take away the print statements for the final push to master to ensure the code is clean.

tests/test_tables.py
```diff
@@ -111,16 +110,19 @@ def test_column(table):
     assert_array_equal(table.column(1), np.array([9, 3, 3, 1]))
 
 
-def test_basic_points(t):
+def test_basic_points(table):
+    t = table
```
davidwagner: What's the purpose behind this change?

deculler:
The use in the 't' and 'u' features is disallowed in the current version of
the unit test framework and provided no simplicity.  The feature needs to
be passed as a parameter to the 'test_...' function.  That 'table' where
't' was used and 'table2' where 'u' was.

Also there were sequential dependence on modifying the test case.  Now each
is standalone.

davidwagner:
We merged #355, which is based on this.  Thanks very much, David!

PR #355, opened 2019-06-09
Cleanup tests

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Fixes test suite as proposed in #354. Also, removes any calls to deprecated API calls from the datascience library in the test suite.


davidwagner:
One test is failing: `ValueError: a and p must have same size` vs `ValueError: 'a' and 'p' must have same size` in the doctest for `sample()`.  See the Travis logs.

davidwagner:
I still don't fully understand why we're renaming the test args from `t` to `table` or why some of those test methods use `t = table` and others use `t = table.copy()`.  However if you tell me that this is correct, I'll rely on you.  The rest looks great.

adnanhemani:
The renaming of the args has to do with the fixtures not being able to be changed during the tests. It makes sense to me, but I'm mainly relying on @deculler's analysis on the t=table vs t=table.copy() (this is also legacy from before - so I'm confident on that as well)

PR #358, opened 2019-06-10
Fix append_column to return the table for method chaining

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Made Table#append_column return the table and updating doctests. Inspired from #344. Thank you @nnj1!


davidwagner:
@adnanhemani , huh, tests seem to be failing here, and I'm not sure what's going wrong.  Some of it is whitespace, but other of it is that values don't seem to be getting printed with dollar signs despite specification of the CurrencyFormatter.  Hope you can investigate -- I can't immediately see what's going wrong.

PR #362, opened 2019-06-11
Update read_table

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Used the solution in #353 and added more robust testing. Thank you @deculler!


adnanhemani:
So far this looks like there's actually no problems - @davidwagner if you could please test against more dense/heavier data files, that'd be great! I have a sinking feeling that something's off here, but I can't tell what and it works 100% (at least locally). Maybe I'm just being paranoid.

davidwagner:
I tested on a few csv files I have lying around, including the large one used in my privacy lecture and some ones that trigger Unicode errors, and the behavior before vs after was identical in all cases.  So this seems fine as far as all my testing indicates.

Issue #367, opened 2019-06-12
Incorrect UserWarning with predicates and strings


davidwagner:
Passing a string to an `are.*` predicate triggers a UserWarning about passing an array or list to a predicate.

It looks like the `check_iterable()` introduced in PR #292 goes overboard; a string is an iterable, so it triggers a warning.  Passing a string ought to be fine.  Perhaps we need to narrow that check to only warn on passing a list or numpy array?  (I doubt anyone is going to pass a Row or a Table so no need to check for those, probably.)

For instance, 

```
> t.where('Sizes',  are.equal_to('L'))
UserWarning: Do not pass an array or list to a predicate.
If you are trying to find rows where two columns are the
same, use table.where('c', are.equal_to, table.column('d'))
instead of table.where('c', are.equal_to(table.column('d'))).
```

(from `datascience/predicates.py`, line 52)

We need to fix this before the summer class starts -- otherwise students will be very confused.

PR #369, opened 2019-06-12
Don't warn when passing a string to a predicate

davidwagner:
[x] Wrote test for feature

**Changes proposed:**
Fixes issue #367, where we incorrectly warned when passing a string to an `are.*` predicate.


davidwagner:
I'm going to optimistically merge this, so we're ready for the summer class.  If we find problems, we can revert it.

Issue #357, opened 2019-06-09
make docs doesn't work


davidwagner:
`make docs` fails without building the docs.  I get some warnings and then an exception:

```
datascience/tables.py:docstring of datascience.tables.Table.apply:7: WARNING: Unexpected indentation.
...
datascience/tables.py:docstring of datascience.tables.Table.pivot:14: WARNING: Definition list ends without a blank line; unexpected unindent.
datascience/tables.py:docstring of datascience.tables.Table.plot:20: WARNING: Definition list ends without a blank line; unexpected unindent.
datascience/tables.py:docstring of datascience.tables.Table.sample:63: WARNING: Unexpected indentation.
datascience/tables.py:docstring of datascience.tables.Table.sample:64: WARNING: Block quote ends without a blank line; unexpected unindent.
datascience/maps.py:docstring of datascience.maps.Circle:20: WARNING: Block quote ends without a blank line; unexpected unindent.
datascience/maps.py:docstring of datascience.maps.Circle:21: WARNING: Definition list ends without a blank line; unexpected unindent.
```

(I think warnings are converted to fatal errors.)  Perhaps this is why the online docs are still back at 0.9.3 even though we're currently at 0.10.8.

davidwagner:
I think the problem is in `readthedocs.yml`.

I figured out how to log into readthedocs.org with my Github account, and can now view the build messages.  I see "Error: Problem parsing YAML configuration. Invalid "python.version": expected one of (2, 2.7, 3, 3.5), got 3.6" (e.g., https://readthedocs.org/projects/datascience/builds/7781403/).  Looking at https://docs.readthedocs.io/en/latest/config-file/v1.html?highlight=3.6 and https://github.com/rtfd/readthedocs.org/issues/4434, it appears we need to add `build: image: latest` to the config file.

Or, perhaps better yet, it seems we should be updating to version 2 style config file for `readthedocs.yml` (https://docs.readthedocs.io/en/latest/config-file/v2.html, https://docs.readthedocs.io/en/latest/config-file/v2.html#migrating-from-v1).

We might want to fix the warnings too.

PR #372, opened 2019-06-13
Fix warnings when building docs

davidwagner:
**Changes proposed:**
Updates the tests to eliminate most of the warnings when running `make docs`, and clean up some outdated/obsolete links.  Partly addresses #357.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -1459,11 +1459,11 @@ class Table(collections.abc.MutableMapping):
         job  | wage
         a    | 10
         a    | 10
-        >>> jobs.sample(k=2, weights=make_array(1, 0, 1, 0)) # Weights must be length of table.
+        >>> jobs.sample(k=2, weights=make_array(1, 0, 1, 0))
         Traceback (most recent call last):
             ...
         ValueError: probabilities do not sum to 1
-        >>> jobs.sample(k=2, weights=make_array(1, 0, 0))
+        >>> jobs.sample(k=2, weights=make_array(1, 0, 0)) # Weights must be length of table.
         Traceback (most recent call last):
             ...
         ValueError: 'a' and 'p' must have same size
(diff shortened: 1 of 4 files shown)


datascience/tables.py
```diff
@@ -1456,12 +1459,10 @@ def sample(self, k=None, with_replacement=True, weights=None):
         job  | wage
         a    | 10
         a    | 10
-        >>> jobs.sample(k=2, weights=make_array(1, 0, 1, 0))
+        >>> jobs.sample(k=2, weights=make_array(1, 0, 1, 0)) # Weights must be length of table.
```
adnanhemani: I think we shouldn't put the comment at the end of this line. Usually comments in doctests where the doctest is failing is meant to be a pointer for a common pitfall - but the error here is different ("probabilities do not sum to 1") than what's in the comment

docs/tutorial.rst
```diff
@@ -363,16 +362,16 @@ Let's do the bootstrap test on the two categories.
...
         """
         Computes one bootstrapped difference in means.
         The table.sample method lets us take random samples.
         We then split according to the number of nonsmokers in the original sample.
         """
         resample = smoker_and_wt.sample(with_replacement = True)
-        bootstrap_diff = resample.column('birthwt')[:num_nonsmokers].mean() - \
-            resample.column('birthwt')[num_nonsmokers:].mean()
+        bootstrap_diff = resample.column('Birth Weight')[:num_nonsmokers].mean() - \
+            resample.column('Birth Weight')[num_nonsmokers:].mean()
```
adnanhemani: I'm not familiar with how to deploy this locally (due to it being a Rust file), but the changes made seem to be 100% correct to make. Let's deploy and verify!

datascience/tables.py
davidwagner: Oh gosh yes, that was dumb!  Fixed.  Thanks.

docs/tutorial.rst
davidwagner: I built the docs with `make docs`, `make serve_docs` and then viewed in my browser at `localhost:8000` and the result looks ok, so I'm guessing this is probably ok.

davidwagner:
This is ready for you to re-review.  Thank you!

Issue #375, opened 2019-06-16
Update documentation with new methods


adnanhemani:
Some new methods that have been merged recently into master do not have a section in the documentation. We should look over the new code checked in and add any new methods to the Rust files that create the documentation

adnanhemani:
Resolved by #377

PR #377, opened 2019-06-16
added new methods into the docs

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
Added the newer methods made into the documentation that gets generated by make docs. As per #375.


adnanhemani:
As part of the review, I would prefer if you could check through recently merged PRs for any new methods that may I have skipped over by accident. Thanks!

davidwagner:
@adnanhemani , Looks good, we now have coverage of all the new methods I could find from the past two years.

PR #378, opened 2019-06-16
make sure that coverage < 5.0.0 for coveralls to work

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
As per https://github.com/coveralls-clients/coveralls-python/issues/203, making sure that coverage version is compatible with coveralls

### DIFF
--- a/linux_environment.yml
+++ b/linux_environment.yml
@@ -1,6 +1,6 @@
 name: datascience
 dependencies:
-- coverage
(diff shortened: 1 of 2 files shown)


adnanhemani:
As per https://travis-ci.org/data-8/datascience/builds/546322118#L2183, this looks like it should work now!

Issue #347, opened 2019-03-07
Table.hist with bins as np.arange does not follow pyplot's x-labels


ketozhang:
See image below

![image](https://user-images.githubusercontent.com/13506487/53926595-ea9cc680-4038-11e9-88b4-86db483c89be.png)

The main issue comes from floating arithmetic representation:

```python
>>> [x for x in np.arange(0.1,1,0.1)]
[0.1, 0.2, 0.30000000000000004, 0.4, 0.5, 0.6, 0.7000000000000001, 0.8, 0.9]
```

But regardless, pyplot has built in way to fix this in `plt.Axes#hist` which isn't represented by the current datascience package:

```python
# Try this
plt.hist(table.column('value'),  bins=bins)

# or this
fig = plt.figure()
ax = plt.axes()
ax.hist(table.column('value'), bins=bins)
```

![image](https://user-images.githubusercontent.com/13506487/53927408-a5c65f00-403b-11e9-84ad-de2f10e99325.png)

I've scanned the source code but couldn't find the issue. The usage of `bins` in  Table#hist is stored in `vargs` in `plt.Axes#hist(**vargs)`

adnanhemani:
From @papajohn in #311:


Numpy v1.14.x (released Jan 6) changed the algorithm (look for the section titled "Many changes to array printing") for the string representation of floating point numbers. If you add this line of code, it changes back to the 1.13.x algorithm, and the histogram x-axis labels are sane (i.e., don't contain the small error terms):

np.set_printoptions(legacy='1.13')

adnanhemani:
I've made a PR at #381 that should take care of this issue. Please review the change by testing it out locally and making sure that it works as expected now. Thanks!

adnanhemani:
Closed by #381.

PR #381, opened 2019-06-18
Hist labels

adnanhemani:
[No] Wrote test for feature

**Changes proposed:**
As per #347. Took Prof. DeNero's advice with adding the legacy printing options for numpy - and this seems to have fixed the issue being caused.


adnanhemani:
Here's the before vs after pictures of the histograms.

Before:
Screen Shot 2019-06-18 at 1 11 26 PM

After:
Screen Shot 2019-06-18 at 1 11 46 PM

However, other than this, I'm not sure how else to test this change and see that this isn't adversely affecting anything else (we don't have testing for histograms). Any ideas? @davidwagner

adnanhemani:
I'm requesting additional reviews for @papajohn and @ketozhang, since they are the originators of the issues to test this fix.

davidwagner:
I'm unsure if this is the best fix.

It should be perfect for the Data 8 class.  From the perspective of the class, this seems like a great fix, as it will fix not only the histogram labels, but make other printed numbers look cleaner.

My one hesitation has to do with whether anyone is using the `datascience` library outside of Data 8.  If they are, `np.set_printoptions()` has global effects.  So, if someone else imports the `datascience` package for anything else, as a side effect it will change the formatting of everything printed using numpy.  Is that OK?  

Should we look into using matplotlib's `set_major_formatter` or `ticklabel_format`?  Or using numpy's `with np.printoptions():` syntax?  Or just go with this fix using `np.set_printoptions()` and see if anyone complains?  What do you think?

adnanhemani:
> I'm unsure if this is the best fix.
> 
> It should be perfect for the Data 8 class. From the perspective of the class, this seems like a great fix, as it will fix not only the histogram labels, but make other printed numbers look cleaner.
> 
> My one hesitation has to do with whether anyone is using the `datascience` library outside of Data 8. If they are, `np.set_printoptions()` has global effects. So, if someone else imports the `datascience` package for anything else, as a side effect it will change the formatting of everything printed using numpy. Is that OK?
> 
> Should we look into using matplotlib's `set_major_formatter` or `ticklabel_format`? Or using numpy's `with np.printoptions():` syntax? Or just go with this fix using `np.set_printoptions()` and see if anyone complains? What do you think?

I thought about this too and thought that because this was the same setting that was present prior to np v1.13, that most everyone would be okay with it going back to that when using our library. But I think you're right - it's better to be defensive. So what I've done is wrapped the entire contents of `draw_hist` with the `with`-syntax (idk why the git diff is being so weird....) so that it uses the prior legacy version when making the histogram.

Did we ever see this with any other type of graph (scatter, boxplot, barplot, or anything else)? If so, I can wrap those easily too, but didn't want to make any changes to those functions prematurely.

Issue #338, opened 2018-10-05
X-axis labels truncated in some cases


davidwagner:
In some cases the x-axis label seems to be truncated.  For example:

```
t = Table().with_columns(
    'Value', make_array(2, 3, 9),
    'Proportion', make_array(0.25, 0.5, 0.25)
)

t.hist(bin_column='Value', bins=np.arange(1.5, 9.6, 1))
```

I don't understand what is causing this.

I have only been able to replicate this with examples that use `bin_column=` (or `count=`), but not with examples that don't.  I don't know if that's fundamental or an accident.

adnanhemani:
Closed by #382.

Issue #337, opened 2018-10-04
Table.hist(bin_column=...) generates wrong x-axis label


davidwagner:
If you pass the `bin_column` parameter to `Table.hist()`, the x-axis is labelled inappropriately.  Consider, e.g.,:

```
t = Table().with_columns(
    'Value', make_array(2, 3, 9),
    'Proportion', make_array(0.25, 0.5, 0.25)
)

t.hist(bin_column='Value', bins=np.arange(1.5, 9.6, 1))
```

This produces the following histogram:

<img src="https://i.postimg.cc/sgPCSKND/hist_wrong_label.png" />

Notice how the x-axis got a label based on `Proportion`.  Instead, the x-axis should be labelled with `Value`.

The bug is in `prepare_hist_with_bin_column()` in `Table.hist()`.  In particular, it generates the x-axis label with the code `w.rstrip(' count')` where here `w` is the label on the column of counts (e.g., `Proportion` in this case).  It seems that the code is expecting/assuming that the label will have a particular format (e.g., that it would be `Value count` in this case).  Instead, the x-axis label should be generated from the column label you passed to `bin_column`.  Thus, it looks like `w.rstrip(' count')` should be changed to `bin_column`.

@papajohn,  @a-adhikari,  do you agree?

adnanhemani:
@davidwagner I do agree about this. I can put up a change quite quickly (I've figured it out, I think) - but do we need the other instructors to agree to the proposed changes?

davidwagner:
I don't think we need to wait for the other instructors.  This shouldn't change anything for existing usage (without bin_column), and for the rare cases with bin_column, it seems clear enough that the current behavior is broken.

adnanhemani:
On second thought, I'm not so sure this is the wrong behavior. I think the original use case for bin_column was for something like a 2-column table where one column is the bins and the other column being the count of a unit as related to the first column. For example, a table like this:

`t = Table().with_columns("bins", np.arange(4), "cars_seen", [6, 4, 5, 9])`

[...]

PR #382, opened 2019-06-18
fixed bug with rstrip

adnanhemani:
[No] Wrote test for feature

**Changes proposed:**
Fixed the bug as described in #338. Was caused as a side-effect of using rstrip which will truncate a set of characters from the back of a string rather than a substring - causing a label like "proportion" to get truncated to "proporti" because "i" was the first letter not in the set of characters " ,c,o,u,n,t".


adnanhemani:
Again, I'm unsure as to how to test this properly.

Untitled.ipynb
```diff
@@ -0,0 +1,116 @@
+{
```
davidwagner: Let's not push a notebook called Untitled.ipynb to the top-level directory.  If this has some tests, let's put it under tests/ with a suitable name.

datascience/tables.py
```diff
@@ -2586,7 +2586,12 @@ def prepare_hist_with_bin_column(bin_column):
             # This code is factored as a function for clarity only.
             weight_columns = [c for c in self.labels if c != bin_column]
             bin_values = self.column(bin_column)
-            values_dict = [(w.rstrip(' count'), (bin_values, self.column(w))) for w in weight_columns]
+            print(bin_column)
```
davidwagner: We shouldn't print things to stdout.  Perhaps the print statements are a remnant of debugging, and need to be deleted?

datascience/tables.py
```diff
@@ -2662,6 +2667,8 @@ def draw_hist(values_dict):
                     x_unit = ' (' + unit + ')' if unit else ''
                     if len(weights) == n:
                         vargs['weights'] = weights[i]
+                    # print(hist_name)
```
davidwagner: Presumably these can be deleted.

davidwagner:
> I'm unsure as to how to test this properly.

If it doesn't mess up any of the histograms in tests/Charts.ipynb, and fixes the problem in #337, I'd say that we have probably tested it sufficiently.

In the long run, we might be able to automate some of the testing for histograms using https://matplotlib.org/devel/testing.html#writing-an-image-comparison-test.

Untitled.ipynb
adnanhemani: Whoops! Thanks for catching!

adnanhemani:
Made all the fixes - thanks for the assist!

Issue #304, opened 2017-11-29
Grouped histograms are missing the x-axis label


papajohn:
`scores.hist('Midterm 1', group='Mentored', bins=np.arange(0, 41, 5), normed=False)`

![download-2](https://user-images.githubusercontent.com/320145/33358063-35cc3274-d47b-11e7-8ff9-02e00ea8313c.png)

adnanhemani:
Closed by #383.

PR #383, opened 2019-06-18
Add label to grouped histogram if only one column plotted

adnanhemani:
[No] Wrote test for feature

**Changes proposed:**
As per #304. I only added support for labels if there is a grouping present and only one column is really being plotted. It doesn't make sense to always add a label to the x-value (in cases where multiple columns are being plotted).


adnanhemani:
Again, no clue on how to test this.

adnanhemani:
So there already is an example in test/Charts.ipynb (it's the cell In[36]). Pushing the other changes though!

adnanhemani:
Ran into a particularly nasty merge conflict. A re-review of all the code might be necessary.

Issue #303, opened 2017-11-29
Inconsistent legends and argument names for grouped graphs


papajohn:
Both args should be called `group` and both legends should have the histogram style.

`scores.scatter('Midterm 1', 'Midterm 2', colors='Mentored')`

![download-1](https://user-images.githubusercontent.com/320145/33358033-f36bb896-d47a-11e7-9840-83687b019549.png)

`scores.hist('Midterm 1', group='Mentored', bins=np.arange(0, 41, 5), normed=False)`

![download-2](https://user-images.githubusercontent.com/320145/33358040-fd0d3a28-d47a-11e7-9765-27ed082dddbf.png)

adnanhemani:
Closed per #384.

PR #384, opened 2019-06-19
Made the legend and arguments identical for hist and scatter

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
As per issue #303. Tested against tests/Charts.ipynb.


Untitled.ipynb
```diff
@@ -0,0 +1,116 @@
+{
```
davidwagner: Looks like this should be deleted or integrated into tests/Charts.ipynb.

adnanhemani: Done!

adnanhemani:
Bumped the version number to v0.12.0 because technically this is a breaking change. Since the break is only for one argument of one function, I decided it's not major enough of a breaking change to upgrade to v1.0.0.

davidwagner:
I went to look at what content we're going to need to convert, and I notice we have about 12 instances of `scatter(..., colors=)` in the lecture demos, about 17 instances in the textbook (lots of overlap with lecture demos), and about 41 instances in assignments.  We're going to need to synchronize those changes with this update to the datascience library.

Also, some old labs, assignments, etc. from prior semesters are no longer going to work with this newer version of the datascience library.  That might be OK but I wanted to get it out there explicitly.

I'm wondering if we should keep around `colors=` as a deprecated synonym for `grouped=` for a little while, to avoid causing breaking changes to existing material.

Apologies for not noticing this before.

davidwagner:
On further reflection, I think we should keep around `colors=` as a backwards-compatible synonym.  Without it the changeover seems too painful (we'd have to synchronize the update in version of datascience on datahub together with a change to the textbook and all assignments etc.).

davidwagner:
@adnanhemani,  I added code to implement the backwards-compatible support for colors=.  Would you look over this and see if it looks good?  If it does, feel free to merge.

adnanhemani:
I think we should add a "Deprecated" warning to users using the "colors" argument and explicitly state that this will be removed by a certain date/version release. It's quite confusing to have both colors and group as arguments around forever. Users should get some time to adapt to the new API call - but shouldn't rely on us supporting colors forever.

datascience/tables.py
```diff
@@ -2350,22 +2355,30 @@ def scatter(self, column_for_x, select=None, overlay=True, fit_line=False,
...
-        if colors is not None:
-            y_labels.remove(self._as_label(colors))
+        if group is not None and colors is not None and group != colors:
+            warnings.warn("Do not pass both colors and group to scatter().")
+        if group is None and colors is not None:
+            # Backward compatibility
+            group = colors
+            # TODO: In a future release, warn that this is deprecated.
+            # Deprecated
+            # warnings.warn("scatter(colors=x) is deprecated. Use scatter(group=x)", FutureWarning)
```
adnanhemani: Let's go ahead and uncomment this. Let's change the warning a bit to say "Table#scatter(colors=x) is deprecated. Use Table#scatter(group=x)" to make it more clear what methods are being referred to.

datascience/tables.py
```diff
@@ -2306,14 +2307,18 @@ def scatter(self, column_for_x, select=None, overlay=True, fit_line=False,
...
 
             ``labels``: A column of text labels to annotate dots.
 
             ``sizes``:  A column of values to set the relative areas of dots.
 
             ``s``: Size of dots. If sizes is also provided, then dots will be
-              in the range 0 to 2 * s.
+                in the range 0 to 2 * s.
+
+            ``colors``: (deprecated) A synonym for ``group``.  Retained
```
adnanhemani: Let's say "Retained temporarily for backwards compatibility. This argument will be removed in future releases.".

adnanhemani:
We could also explicitly catch the "colors" argument via the **vargs argument into the function and output an error message to use the group argument instead as well?

datascience/tables.py
davidwagner: I'd prefer to leave this for a future release, so we can have the summer class adopt this updated version of the datascience library without triggering alarming-looking warnings.

datascience/tables.py
davidwagner: Oh, I like that.  Done.

davidwagner:
I agree about warning about deprecation.  I added that.  I would prefer to avoid triggering warnings just yet, so that we can deploy this updated version of the datascience to datahub (without triggering unsightly warnings for students), then update all of the Data 8 git repos; then once we've done that, let's turn on the warnings.  I don't know how to do it with **vargs -- I confess I'm a bit clueless how vargs works, or whether that's better than an explicitly named colors= argument, so I'll rely on your expertise there.

adnanhemani:
I see - that sounds good. Let's merge this as is - and then start working on renaming everything in our course materials. For the future release, I think I have a clean solution to continue warning not to use colors. I'll merge this first and then create a new PR with those changes that we can leave open until we think it's a good time to merge the newer changes.

PR #385, opened 2019-06-24
Maps now completely working with latest version of Folium (v0.9.1)

### LINKED ISSUE #346: Upgrade folium dependency to >=0.6.0
The folium dependency is currently [pinned to 0.2.1](https://github.com/data-8/datascience/blob/master/requirements.txt) (released >2.5 years ago), while the current release of folium is 0.7.0.

There are many interesting features in the more recent [releases](https://github.com/python-visualization/folium/releases) since 0.2.1. However, the API has changed significantly, which cause the datascience [maps](https://github.com/data-8/datascience/blob/master/datascience/maps.py) functionality to break.

adnanhemani:
[X] Wrote test for feature
[No] Added changes in the Changelog section in README.md
[No] Bumped version number (will likely need to be v1.0.0)

**Changes proposed:**
Has all the required changed to support the Maps module with the latest version of Folium (v0.9.1).

Closes #346.


adnanhemani:
There's a lot to go over with these changes, here's the abridged version:
* Both the code and the testing files are very unorganized and messy. This is expected behavior at the moment - the testing files in the root directory are there because of a development environment issue my computer has currently. I will fix the issue as soon as possible - but can't at the moment due to the active development. **Please do not remove the file at the root directory - I will do that before a final merge to master, but need those to be around at the moment for my workflow.**
* There's been a regression in the test_tables.py tests locally (at least for me) - it's the same issue where we're testing for a specific error message. Maybe this has to do with my development environment issues, but I'm really not sure.

requirements.txt
```diff
@@ -1,3 +1,3 @@
-folium==0.2.1
+folium==0.9.1
```
davidwagner: I'm inclined to suggest `folium>=0.9.1` (if we later discover that there is some known problem with a future version of folium we can change this to pin a particular version).  We might pin `folium==0.9.1` on datahub but I'm not sure it is necessary or desirable in `requirements.txt`.

davidwagner:
This is looking great!

I don't see the test_tables.py regression when I run `make test` on this branch.

The folium repository seems to be somewhat low activity: last activity was a month ago.  Let's hope this gets fixed in the not-too-distant future.  In the meantime, `folium==0.8.3` seems to work, so maybe we switch to that version of folium for now, and update to the latest folium once it fixes the bug you reported.

For documentation, I'd be inclined to leave that to a future pull request.  I think we could handle that on an on-demand basis, i.e., if a class is going to use this module more, then document the parts they're going to use, and maybe ask the current users of folium what they'd want.  But I'd separate that out into a separate future pull request, as it's a separate feature.  We can still merge this one and it'll be sufficient for current Data 8 usage without any new documentation.

adnanhemani:
Sounds good about the documentation - I'll create an issue for better documentation and then we can write those sections as needed.

I've made a PR at Folium for the change. It's a relatively small change and I've been told that they can release v0.9.2 after the changes are submitted. Per this, it might be best to wait for the changes to get merged into Folium - and then change the requirements.txt to say folium>0.9.3.

davidwagner:
I don't see any reason to wait.  It's easy enough to set `folium==0.8.3` now and then do another pull request later that changes `requirements.txt` to `folium>=0.9.2` (and does nothing else).

adnanhemani:
I haven't tested at v0.8.3 with all the notebooks (I'm especially using the Geospatial connector's notebooks for this). I'm sure the API hasn't changed much between those versions but I'm still uncomfortable rolling the folium version number back without further testing (which will also require some time). I'm still also not done cleaning up the code - sorry for the delay!

adnanhemani:
@davidwagner I think this code is good to go with the current folium version. I'll start working on testing compatibility with folium v0.8.3 (but to be completely honest, the warning is very small - and can only be triggered incorrectly with 2 certain argument options - I'm fine with merging with just folium version 0.9.2 for now and keeping an active eye on the folium repo if you are too). The testing for v0.8.3 might take a few days though :(

I will probably need a stylistic- and architectural- review, if possible. Thank you!

adnanhemani:
The code as-is does not work completely with folium v0.8.3. This is due to a problem with clustered markers. @davidwagner - what are your thoughts?

I don't think that merging now is much of a problem - the warning is _very_ rare.

davidwagner:
Oh!  I didn't realize that the impact of upgrading to folium 0.9.1 would be so low.

I did a quick review of Data 8 course notes to see if we're using the colors lightgray or black anywhere in the course materials.  Project 1 uses `'black'` for a map with `Circle.map_table(scaled)`; will that be affected?  Other than that, I couldn't find any other place that uses those two colors.  So, if Project 1 still works with this update & folium 0.9.1, let's merge this; otherwise, let's hold off, as I don't want to break Project 1.  Does that sound good?

adnanhemani:
So Project 1’s map will still run (the black marker will show up with the right color), but it will produce the wrong warning (non-fatal). If my memory is correct, students aren’t typing the code in Proj 1 for the map, so I don’t think the warning will alarm students much and the map will still be correct. 

Now that I think about it, I think I can suppress that particular warning. If you think it’s necessary, I can try to suppress the warning for the time being and then we can remove the suppression and up the Folium version number when the new version releases.

@davidwagner, which do you prefer? Suppress the warning - or just let it go?

davidwagner:
Let's go for it.

PR #389, opened 2019-06-28
removed colors argument for Table#scatter

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
The next commit to remove support for the colors argument altogether and warn the user to use group instead.


datascience/tables.py
```diff
@@ -2356,14 +2355,8 @@ def scatter(self, column_for_x, select=None, overlay=True, fit_line=False,
...
         x_data, y_labels =  self._split_column_and_labels(column_for_x)
-        if group is not None and colors is not None and group != colors:
-            warnings.warn("Do not pass both colors and group to scatter().")
-        if group is None and colors is not None:
-            # Backward compatibility
-            group = colors
-            # TODO: In a future release, warn that this is deprecated.
-            # Deprecated
-            # warnings.warn("scatter(colors=x) is deprecated. Use scatter(group=x)", FutureWarning)
+        if "colors" in vargs and vargs["colors"]:
```
adnanhemani: Now that I think about it, this should warn on run of the code instead of just the first run. Thoughts?

deculler:
The previous handling of colors for scatter was rather broken because of
internal handling through a shared wrapper.  If you specified the kwarg it
errred.  So no existing code is passing the colors= kwarg in.  But lots of
code wishes it could.  The old use of colors was to provide difference
colors for different columns in the scatter.  That behavior is reasonable
when no color arg is provided.  The other (more typical) use is to use
distinct colors for distinct elements in a (y) column.  A very natural
idiom would be to have a "color column" or array.  The challenge here is if
one wants different coloring for each of more than one y-column in the
table.  We currently have the y-columns as the set of non-x columns.  The
color kwarg should be conformant with that.  With a single y-column it
could be an array.  More generally it could be a sequence of column labels,
which would be removed form the y-column set and utilized in order to color
the corresponding columns.  Slight generalization would be to also accept a
"color table" that would be conformant with the y-columns of the scatter.

David E. Culler
Interim Dean of Data Sciences
Friesen Professor of Electrical Engineering and Computer Sciences

davidwagner:
@adnanhemani,  Right now we accept either `group=` or `colors=` (as of the latest version).
Next I'd like to deploy that, update all the course materials, and let that bake for a few days.
Then, I'd like to update so we accept either `group=` or `colors=`, but warn on use of `colors=`, and deploy that for a little while.
Then, after that's been out for a little while, we can remove support for `colors=` and only support `group=`.
I'd prefer to go through the intermediate step first before we merge this one.
So let's keep this one on hold until we've done those steps.  Thanks for preparing this, it will be useful to us in the future!

davidwagner:
@deculler, we do have many instances of `Table.scatter(..., colors=...)` in the Data 8 course materials: see https://github.com/data-8/datascience/pull/384#issuecomment-504199174.

Your idea about supporting choosing a different color for each point with a colors column is a neat one.  That sounds like something we can handle as a separate feature request.  Want to file an issue for that feature request so we can keep track of it?  We'll probably prioritize based on whether we have anyone who wants to use it.

datascience/tables.py
davidwagner: Agreed.  Seems like warning every time might be preferable.

deculler:
That is what the colors= kwarg in matplotlib scatter does.  Happy to do it
separately.  Would have written the code a long time ago, but the internal
generation of a color= thru _draw was a pain to back out.

adnanhemani:
@davidwagner maybe we should look into merging this through at some point? I think all the class materials were changed a while back to reflect the new method signature?

davidwagner (approved):

PR #392, opened 2019-07-01
Shallow select

### LINKED ISSUE #265: Allow shallow copy on select
Right now, `select` always makes a copy of column data. There are probably many cases in which it shouldn't. Adding a `copy_values=True` optional arg should do the trick.

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Closes #265.

Allows for users to "shallow" copy data from the table in `Table#select`. Sets the default behavior to not shallow-ly copy, as is normal behavior right now.


davidwagner:
I'll leave whether to merge this up to @papajohn -- I'll await his response on #265.

PR #393, opened 2019-07-01
Supporting new take syntaxes

### LINKED ISSUE #332: Table.take should accept a variable number of arguments
Stacking parentheses and brackets is confusing for beginners (e.g., `make_array(1, 2, 3)` is easier to understand than`np.array([1, 2, 3])`). Taking multiple rows with `Table.take` should also accept this syntax, so that all of the following work and produce the same results:
```
t.take(np.arange(3))
t.take[:3]
t.take([0, 1, 2, 3])
t.take(0, 1, 2, 3)
```

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
This allows for another syntax to be used with `Table#take` and `Table#exclude` (for example, `t.take(0, 1, 2, 3)` or `t.exclude(2, 3)`). As per #332.

Closes #332.


datascience/tables.py
```diff
@@ -3066,11 +3072,20 @@ def __getitem__(self, row_indices_or_slice):
...
+        >>> grades.take(0, 2)
+        letter grade | gpa
+        A+           | 4
+        A-           | 3.7
         >>> grades.take(10)
         Traceback (most recent call last):
             ...
         IndexError: index 10 is out of bounds for axis 0 with size 6
         """
+        # print(row_indices_or_slice)
```
davidwagner: These look like remnants of personal debugging, and should be removed before merging.

davidwagner:
Can we add an example to the docs for `exclude()`, too?

Otherwise, looks great!

adnanhemani:
Made the changes, thanks!

PR #394, opened 2019-07-02
New optional argument for make_array

### LINKED ISSUE #339: Add dtype options to make_array
We had an issue recently in our class where students using windows machines have default int32 types, so allowing for a dtype argument to be passed to the numpy function would be good.

image

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Adds a new optional argument to specify the dtype of the array that we're creating. As per #339.

Closes #339.


davidwagner:
While this is a reasonable implementation of the requested feature, I'm not enthusiastic about the proposed feature, and would like to find a better way of addressing the problem.  I left comments on issue #339.  Sorry to have wasted your time on this.

adnanhemani:
Here are the new changes - it's checking that if every argument to `make_array` is an integer, it will automatically pass into np to make the array of type int64.

PR #398, opened 2019-07-03
Changed default formatting back to without commas

adnanhemani:
[X] Wrote test for feature
[No] Added changes in the Changelog section in README.md
[No] Bumped version number (delete if unneeded)

**Changes proposed:**
As per the email thread with the Su19 instructors. This should change the default formatting of an integer column back to no commas - as it was prior to #355 (the comma formatting can be set again using the NumberFormatter). No version number bump as this is reverting the behavior back to the original settings.


davidwagner:
Looks good!  I bumped the version number, so we can specify a particular version with this change.

PR #402, opened 2019-07-11
Adds exception for Table.with_columns(...)

### LINKED ISSUE #333: Should detect Table.with_columns() and warn the user
A common error that I see from students is to write code like `Table.with_columns('x', arr)`, when they should have used `Table().with_columns('x', arr)`.  I suggest we try to detect that.

I suspect we could add a check at the start of the `with_column()` method to issue an error if `type(self)==str`, or maybe more broadly if `type(self)!=Table` (as I don't think we expect anyone to subclass Table).

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Closes #333.


setup.py
```diff
@@ -11,7 +11,7 @@
 
 tests_requires = [
     'pytest',
-    'coverage==4.5.3',
+    'coverage==3.6.0',
```
davidwagner: What's the reasoning behind this change?

adnanhemani: Yup, sorry...probably forgot to switch branches before pushing

adnanhemani:
Changes made!

PR #408, opened 2019-08-13
Made Marker.map_table more flexible

### LINKED ISSUE #406: Marker.map_table not as flexible as expected
You cannot add more attributes to a Table's rows for individual features as is in Marker.map_table, due to the default arguments in Marker.map. We should make the Marker.map_table method more robust so that it properly handles the default arguments and bypasses them, if necessary.

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
Allows for user-defined options to be made as part of `Marker.map_table`. Closes #406.


davidwagner:
Sorry that I didn't see this earlier!  This looks good to me.

PR #411, opened 2019-09-15
Fix shaded hist bounds

rameshvs:
[x] Wrote test for feature

**Changes proposed:**
* Fix shaded histograms so that bounds are computed properly when using 0 as the left/right end
* Fix computation when edges overlap bin boundaries
* Add tests for shaded histogram bounds computations


rameshvs:
I was able to pass all tests locally, so I'm going to go ahead and merge since it seems like there's some kind of environment issue with travis.

PR #412, opened 2019-09-16
Remove libgfortran dependency

davidwagner:
**Changes proposed:**
Builds on Travis were failing because of a dependency on installing `libgfortran` via conda.  I'm not sure that we ever used that for anything, so I removed it; everything still seems to work.

### DIFF
--- a/linux_environment.yml
+++ b/linux_environment.yml
@@ -8,7 +8,6 @@ dependencies:
 - freetype
 - ipython
 - ipython_genutils
-- libgfortran
 - libpng
 - libxml2
 - matplotlib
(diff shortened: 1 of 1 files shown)


davidwagner:
@adnanhemani , could you review this to check that I'm not doing something stupid or that there's some reason I'm missing why we need a libgfortran dependency?  Thanks!

adnanhemani:
So I looked through the Git blame...seems like @SamLau95 add it in a commit for the Linux environment (seems like it was done on purpose). I'm not sure of any reason why we'd need it (I don't see any code that would be dependent on it) - but I'm guessing only @SamLau95 can tell us why it was added?

SamLau95:
Sadly I don't remember why it was added. Chances are it was autogenerated from `conda`. If it makes the build work, then I say let's remove it.

PR #413, opened 2019-09-16
fixed bug from previous version

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
Fixed a bug and cleaned up code inadvertently introduced by v0.15.1. The issue related to the fact that slicing NumPy arrays only provides a view into the original array rather than creating a copy. This caused the overall bins to get written every time the particular line of code was run. Also cleans up print statements that were probably left behind from debugging.


datascience/tables.py
```diff
@@ -2638,9 +2638,15 @@ def prepare_hist_with_group(group):
         values_dict = collections.OrderedDict(values_dict)
         if left_end is not None or right_end is not None:
             if left_end is None:
-                left_end = min([min(self.column(k)) for k in self.labels if np.issubdtype(self.column(k).dtype, np.number)])
+                if bins[0]:
```
adnanhemani: These new checks are for making sure that if the user specifies bins, then the histogram shading goes all the way to the edge of the first/last bin

Issue #416, opened 2020-03-23
group with `nan` values is strange


yifanwu:
If you load the following csv file and attempt to do group by on county, it treats each `nan` as unique---I think the nans should be treated as the same value. Happy to be convinced otherwise.

```
county,price,year
,17899,2012.0
,0,2016.0
,46463,2015.0
,0,2016.0
,49999,2018.0
,13999,2009.0
,34500,2017.0
```

adnanhemani:
Hi @yifanwu. I took a look - it's reproducible (thank you for the example), here's why: np.float64 NaN objects aren't equal to each other in Python. Here's a SO post with a deeper explanation: https://stackoverflow.com/questions/6441857/nans-as-key-in-dictionaries

That being said, I agree that this isn't the best behavior - I'm going to be putting in a PR for this. Thanks for reporting the issue!

adnanhemani:
Closed by #417. Hope this helps :)

yifanwu:
Thanks Adnan for the timely fix!

PR #417, opened 2020-03-29
Fixes behavior for bug In grouping where np.float64 objects exist

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
As mentioned in #416, there's a bug that occurs if there's NaNs of type np.float while grouping (this error does not occur if the NaN is of type str). This is a fix for solving this bug in case of both `Table#group` and `Table#groups`.


adnanhemani:
Seems like there's an issue with our Travis build again... :(

Will look into that next.

adnanhemani:
The remaining failing test is unrelated to this PR: will be raising another PR shortly to fix it.

PR #430, opened 2020-06-12
docs: Added more details to some docstrings

savithrikb:
Added documentation for methods from_records, _with_columns, and
from_array in tables.py

[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -95,7 +95,7 @@ class Table(collections.abc.MutableMapping):
            Returns:
 
                If the list is empty, it will return an empty table.
-               Otherwise, it will return a table with the labels as the column name, and the corresponding data.
+               Otherwise, it will return a table with the dictionary's keys as the column name, and the corresponding data.
                If the dictionaries do not have identical keys, the keys of the first dictionary in the list is used.
 
            """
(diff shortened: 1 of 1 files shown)


datascience/tables.py
```diff
@@ -126,7 +138,16 @@ def read_table(cls, filepath_or_buffer, *args, **vargs):
         return cls.from_df(df)
 
     def _with_columns(self, columns):
-        """Create a table from a sequence of columns, copying column labels."""
+        """Create a table from a sequence of columns, copying column labels.
```
adnanhemani, line 141: As this function is using the Python convention for a private method, I don't think we should be adding any further documentation for this method. It would give the wrong impression to end users that this method can/should be used by them (and this method is simple enough for developers to understand on a quick skim)

datascience/tables.py
```diff
@@ -86,7 +86,19 @@ def from_rows(cls, rows, labels):
...
+        """Create a table from a sequence of records (dicts with fixed keys).
+
+           Args:
+
+               records: A list of dictionaries with same keys.
+
+           Returns:
+
+               If the list is empty, it will return an empty table.
+               Otherwise, it will return a table with the labels as the column name, and the corresponding data.
```
adnanhemani, line 98: Replace "labels" with "the dictionary's keys" (labels and columns are synonyms in this library)

datascience/tables.py
savithrikb: Done. Thank you for your response.

datascience/tables.py
savithrikb: Done. Thank you for your response.

PR #432, opened 2020-06-15
Include ipynb files in tests and coverage generation

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**

Including ipynb file to be run as part of the Travis CI build.


adnanhemani:
This should be good to go - the failures are happening as a result of the fix in #431 not being merged. When that goes through, this should be good to go! :)

davidwagner:
Is Charts.py used somewhere?  Wanted to make sure it is a useful file and is not a leftover from a test or debugging or something.

adnanhemani:
Oof - no you're right, nice catch! (Was an artefact left over from testing the process locally 😢 )

Pushing again without it

adnanhemani:
hmm this didn't do exactly what I expected...the coverage didn't go up which tells me maybe I've made an error with the coverage --append. I'll comment back when I get this sorted out

adnanhemani:
💥 it's done! I've changed us off of using Coverage.py and instead using PyTest's in-built coverage tool. The new test coverage using the ipynb notebooks for some reason wasn't working well with the existing Coverage.py infrastructure. 

@davidwagner could you please have a look when you can? When this gets pushed into master, I'll probably look at increasing our test coverage as far as possible :)

davidwagner:
@adnanhemani   Sorry to be so slow getting back to you on this.  How do I test this?  I tried running `make test` but I get the following error:

```
ERROR: usage: setup.py [options] [file_or_dir] [file_or_dir] [...]
setup.py: error: unrecognized arguments: --nbval-lax --cov=datascience
  inifile: None
  rootdir: /home/daw/datascience/datascience
```

I did a `pip install` of `nbval` just in case that was the issue, but it didn't solve the problem.

davidwagner:
The latest build on Travis for this branch seems to have failed: https://travis-ci.org/github/data-8/datascience/builds/700518382  I am not sure why that build failed, when another build just 15 minutes before passed (https://travis-ci.org/github/data-8/datascience/builds/700516647).  It looks like the only thing that changed was updating from folium 0.2.1 to folium 0.9.1.

adnanhemani:
So I think I fixed the issue via the new code I'm pushing. But also, running `pip install nbval pytest-cov` should fix it on your local system. Or download the new version of the code and then run `conda env update -f osx_environment.yml` :) Pushing fix in a second.

davidwagner (approved):

davidwagner:
For some reason I still can't do `make test` on one machine (Linux) but can on another, even after installing all the dependencies.  I don't know why, but I'm going to merge this because it's great stuff!

PR #441, opened 2020-08-13
Interactivity Updates for Tables and Maps

chrispyles:
[x] Wrote test for feature

**Changes proposed:**
This PR includes new interactivity updates that allow for plotly-based plots in addition to their matplotlib defaults and new options for `Marker#map_table` and `Circle#map_table`. It also includes a Data 8 student-friendly reference notebook in the documentation.


chrispyles:
Hey @davidwagner @adnanhemani this PR is failing due to the coverage decrease. Because all of the methods added aren't tested by pytest (since they're all plotting/mapping) what do you suggest be done so this check passes?

davidwagner:
Adnan wrote some code to include everything covered by notebooks in the coverage numbers (PR #432), so that might be one clean solution, but I haven't figured out how to get that PR working on my end yet.

davidwagner:
It looks like this introduces a bunch of new functionality: e.g., clusters, index maps, colorbar scales, radius units.  I'm guessing those are orthogonal to whether the maps are interactive vs not.  Can you talk a bit about that?  What's the motivation for adding that functionality?  It looks like could use some careful attention and I'm not sure I know this codebase well enough to render a thoughtful opinion on them.  If we introduce these arguments, will we have to continue to support them in the non-interactive plot methods, and should we expect that to impose a maintenance burden?

I also see a bunch of additional documentation, which looks useful independent of any interactivity changes.  Cool.

davidwagner:
@adnanhemani Would you like the opportunity to review this PR?

datascience/tables.py
```diff
@@ -2260,6 +2331,16 @@ def plot(self, column_for_xticks=None, select=None, overlay=True, width=6, heigh
         >>> table.plot('days', 'price') # doctest: +SKIP
         <line graph with days as x-axis and line for price>
         """
+        global _INTERACTIVE_PLOTS
+        if _INTERACTIVE_PLOTS:
+            return self.iplot(column_for_xticks, select, overlay, width, height, **vargs)
+
+        if width is None:
```
adnanhemani, line 2338: No need for this. The _visualize method already has default params for this, unless I'm missing something...

datascience/tables.py
```diff
@@ -2398,6 +2590,15 @@ def barh(self, column_for_categories=None, select=None, overlay=True, width=6, *
         >>> furniture_table.barh('Furniture', make_array(1, 2)) # doctest: +SKIP
         <bar graph with furniture as categories and bars for count and price>
         """
+        global _INTERACTIVE_PLOTS
+        if _INTERACTIVE_PLOTS:
+            # If width not specified, default width originally set to 6,
+            # Multiply by 96 assuming 96 dpi
+            return self.ibarh(column_for_categories, select, overlay, width, **vargs)
+
+        if width is None:
```
adnanhemani, line 2599: also not necessary

datascience/tables.py
```diff
@@ -2534,6 +2879,29 @@ def scatter(self, column_for_x, select=None, overlay=True, fit_line=False,
...
+                labels = labels,
+                sizes = sizes,
+                s = s / 4, # Plotly dot sizes are much smaller, so divide s by 4
+                width =  width,
+                height = height,
+                colors = colors,
+                **vargs
+            )
+
+        if width is None:
```
adnanhemani, line 2899: Not necessary

datascience/tables.py
```diff
@@ -2746,6 +4172,44 @@ def hist(self, *columns, overlay=True, bins=None, bin_column=None, unit=None, co
...
+                side_by_side = side_by_side,
+                left_end = left_end,
+                right_end = right_end,
+                width = width,
+                height = height,
+                rug = rug,
+                **vargs
+            )
+
+        if width == None:
```
adnanhemani, line 4207: Not necessary

docs/tables.rst
```diff
@@ -137,12 +137,20 @@ Visualizations
     :toctree: _autosummary
 
     Table.plot
+    Table.iplot
```
adnanhemani, line 140: If our preferred way of users using the interactive plots is by using the interactive method, then I'm not sure that we should be documenting the methods straight. Rather, merge this documentation into the non-interactive method one's - and that shouldn't add much extra stuff, given that the method signatures should be identical.

adnanhemani:
Making some comments here since diff is too large to show:
* The URL for world-countries.json needs to be changed back to the link. It's been made a local filename for some reason and it shouldn't be.
* When looking at plotly-charts.ipynb, under the `Working with Returned Figures` header, all the Figures are getting printed out. We need to suppress those.
* All the CSVs that were added in for the reference-notebook need to be turned into reading from URLs (I'm guessing you downloaded these from somewhere in CSV format). These files added an extra ~12MB to the entire repo and we try to keep the repo as light as possible. I'd say the same for the geocoding CSV, but that's still understandable as we want all our functionality to work offline, as much as possible.

Overall, to respond with my two cents to @davidwagner's thoughts...I think there's a bunch of new functionality and complication that we are introducing via this PR - but after reading through it all, I'm leaning on the side of saying it's worth it to have in our package. There will no doubt be extra maintenance overhead (heck, there's so much code for the interactive plots that I just skipped because I don't have the time nor mental capacity to digest all of it at the moment) but I think it can be manageable if @khsu2000 and @chrispyles can commit to taking up maintenance for all interactive plotting methods; I think I understand most of the new maps functionality to take responsibility for it if all the changes I asked for are made.

datascience/tables.py
chrispyles: I think all of these are necessary... `Table#_visualize` specifies the defaults of the width and height as default values for kwargs but since it is called with these kwargs set to the values in the function (i.e. `self._visualize(..., width=width, height=height, ...)`) these get overwritten with `None`. Will put these into a class attribute so that we don't hardcode the same values multiple times but I think these statements are necessary.

datascience/tables.py
```diff
@@ -18,12 +18,16 @@
 import matplotlib.pyplot as plt
 import pandas
 import IPython
+import plotly.graph_objects as go
```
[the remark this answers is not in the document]
chrispyles: fixed. Created `Table#_import_plotly` which gets called in all the right places.

docs/tables.rst
chrispyles: changed.

datascience/tables.py
adnanhemani: ahh you're right, I was thinking of something else. good catch!

chrispyles:
@adnanhemani @davidwagner OK I've pushed resolutions to some of these comments. Resolved the specific edits above and addressed all 3 of @adnanhemani's bullet-points above. @khsu2000 is working on the updates to maps, which we anticipate having completed by Friday.

datascience/maps.py
```diff
@@ -557,16 +690,75 @@ def map_table(cls, table, clustered_marker=False, **kwargs):
...
                 areas = this_col
-            else:
+            elif col not in excluded:
                 other_attrs[col] = this_col
+
+        if "cluster_by" in table.labels:
+            clustered_marker = True
+            cluster_column = table.column("cluster_by")
+            cluster_labels = list(set(cluster_column))
+            table_df = table.to_df()
```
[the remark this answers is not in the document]
khsu2000: Done, rewrote this using datascience tables only.

datascience/maps.py
```diff
@@ -701,3 +905,77 @@ def _lat_lons_from_geojson(s):
         return [(lat, lon)]
     else:
         return [lat_lon for sub in s for lat_lon in _lat_lons_from_geojson(sub)]
+
+
+def get_coordinates(table, replace_columns=False, remove_nans=False):
```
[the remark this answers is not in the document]
khsu2000: Done, rewrote this using datascience tables only.

datascience/maps.py
```diff
@@ -80,6 +84,18 @@ def __init__(self, features=(), ids=(), width=960, height=500, **kwargs):
         tile_style = None
         if "tiles" in kwargs:
             tile_style = kwargs.pop("tiles")
+        self._index_map = self._cluster_labels = self._colorbar_scale = None
+        self._radius_in_meters = False
```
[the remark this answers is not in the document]
khsu2000: Used self._attrs for these args instead, so that there's no new if statements in Map initializer.

datascience/maps.py
```diff
@@ -112,15 +128,73 @@ def __iter__(self):
     def _set_folium_map(self):
         self._folium_map = self._create_map()
         if 'clustered_marker' in self._attrs and self._attrs['clustered_marker']:
-            marker_cluster = MarkerCluster().add_to(self._folium_map)
+            def customize_marker_cluster(color, label):
```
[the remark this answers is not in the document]
khsu2000: This is for a new cluster_by feature that can take groups of markers and assign them each to a separate marker_cluster. In order to distinguish the groups, we needed to make new marker_cluster icons that include a mouse-over feature.

chrispyles:
Added tests to get the coverage back to where it was @adnanhemani. @khsu2000 and I are meeting tomorrow to discuss the last edit but all else is gtg.

datascience/maps.py
```diff
@@ -506,6 +615,18 @@ def map(cls, latitudes, longitudes, labels=None, colors=None, areas=None, other_
         assert len(latitudes) == len(longitudes)
         assert areas is None or hasattr(cls, '_has_radius'), "A " + cls.__name__ + " has no radius"
         inputs = [latitudes, longitudes]
+        index_map = include_color_scale_outliers = cluster_labels = colorbar_scale = None
+        radius_in_meters = False
```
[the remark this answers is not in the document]
chrispyles: OK we've collapsed some of this information into a new set of kwargs called `map_kwargs` which gets passed into the `Map` constructor. @khsu2000 has some objections to passing around the original `kwargs` dictionary that he can elaborate on if needed.

chrispyles:
OK @adnanhemani @davidwagner all changes have been made, tests added, etc. Should be ready to merge from our end; please let @khsu2000 and I know if anything else is needed.

Also re: maintenance, @khsu2000 and I are OK with taking responsibility for maintenance of the new plotly stuff and whatever else as needed.

adnanhemani:
Hi @khsu2000 and @chrispyles - last thing on my end and then it looks great to me. @davidwagner would you also like to review to this make sure I haven't left anything out?

datascience/maps.py
```diff
@@ -707,3 +905,79 @@ def _lat_lons_from_geojson(s):
...
+    Args:
+        table: A table with counties that need to mapped to coordinates
+        replace_columns: A boolean that indicates if "county", "city", "state", and "zip code" columns should be removed 
+        remove_nans: A boolean that indicates if columns with invalid longitudes and latitudes should be removed
+        
+    Returns:
+        Table with latitude and longitude coordinates 
+    """
+    assert "zip code" in table.labels or (("city" in table.labels or "county" in table.labels) and "state" in table.labels)
+    ref = Table.from_df(pandas.read_csv(pkg_resources.resource_filename(__name__, "geodata/geocode_states.csv")))
```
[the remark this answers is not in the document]
chrispyles: Just pushed a fix. `pkg_resources.resource_filename` basically allows us to find the path to a data file that is stored in the same location where the package is installed, so we can access the data file we need to construct the table.

chrispyles:
Removed the `colors` argument from `Table#iscatter`. Let's see if travis passes.

adnanhemani (approved):

adnanhemani:
LGTM - I'll let @davidwagner have final say before I merge this since it is quite large.

davidwagner:
@adnanhemani   This looks good to me!  I have a small pending comment (about listing plotly twice in linux_environment.yml), but that should be easy to fix.  I think we're good to go to merge it.

Thank you for all your work on this, @chrispyles and @khsu2000!

davidwagner:
I've merged this -- thank you for all the work, @chrispyles and @khsu2000!

I will rely on you in the future for any software maintenance or bug reports that touch these pieces.

PR #448, opened 2020-08-31
Issue 444: Fixes tutorials DeprecationWarning

### LINKED ISSUE #444: Avoid Table.with_columns([...]) in tutorial.rst
`docs/tutorial.rst` uses syntax of the form `Table.with_columns([...])` on line 390.  This is unnecessarily ugly; cleaner is `Table.with_columns(...)` without creating a list.

Also, it breaks building docs, with the following error:

```
Warning in /home/docs/checkouts/readthedocs.org/user_builds/datascience/checkouts/master/docs/tutorial.rst at block ending on line 229
Specify :okwarning: as an option in the ipython:: block to suppress this message
----------------------------------------------------------------------------
../datascience/tables.py:630: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  values = np.array(tuple(values))
```

Rather than trying to suppress the warning, I think we should just change the tutorial to use better style.

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
Fixes the tutorials deprecation warning.

Closes #444

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -60,11 +60,11 @@ columns.
 
 .. ipython:: python
 
-    t = Table().with_columns([
+    t = Table().with_columns(
         'letter', ['a', 'b', 'c', 'z'],
         'count',  [  9,   3,   3,   1],
         'points', [  1,   2,   2,  10],
-    ])
+    )
 
     print(t)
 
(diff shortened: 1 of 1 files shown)


adnanhemani:
Closes #444.

davidwagner (approved):

Issue #163, opened 2015-12-21
Methods that need more docs


SamLau95:
Housekeeping for which methods still need docs (just in the order they appear in
the documentation, not by priority).

These methods have just a sentence or two in the docstring.

**Creation**
- [x] from_rows (adnanhemani edit: this method is now deprecated. No docs required.)
- [x] from_records
- [x] read_table
- [x] from_df
- [x] from_array

**Accessing Values**
- [x] num_rows
- [x] rows
- [x] column_labels
- [x] columns
- [x] column_index
- [x] first
- [x] last

**Mutation**
- [x] set_format
- [x] move_to_start
- [x] move_to_end
- [x] append
- [x] remove

**Transformation (I'd say these are the most used methods)**
- [x] copy
- [x] drop
- [x] where
- [x] sort
- [x] group
- [x] groups
- [x] stack
- [x] join
- [x] stats
- [x] move_column
- [x] pivot_bin (needs examples)

**Exporting/Displaying**
- [x] show
- [x] as_text
- [x] as_html
- [x] #520
- [x] to_df
- [x] to_array

**Visualizations**
- [x] plot
- [x] points

**Others**
- [x] Row class
- [x] Rows class

---

These methods have better docs (eg. an example or a description of the args) but
I'd still say they could use more documentation.
- [x] from_columns_dict
- [x] pivot
- [x] pivot_bin

Brainz456:
Any of these still require updating?
I would be interesting in writing some up.

adnanhemani:
@Brainz456 - yes these do. Please check the latest documentation at data8.org/datascience for the latest state/version of the documentation. If you do decide to work on this, please update the checklist above as well as you create PR(s) for these methods. Thanks!!

erichamers:
Hey guys! Is anyone still working on this issue? I don't see any assignees, I'd be willing to jump in and help with this.

adityakuppa26:
@SamLau95  @adnanhemani  Isn't column_labels deprecated? Do we still add docstrings for such?

adityakuppa26:
#440

adityakuppa26:
issue 163: added documentation for all accessing value methods #440

adnanhemani:
@erichamers - we don't have any one person solely working on this issue. If you can, please fork our repo, and submit a PR! It'd be super appreciated!

[...]

PR #457, opened 2020-09-13
Add docs & tests for Table.set_format

hmstepanek:
- [x] Add docs & tests for Table.set_format

This is a partial fix for [Add documentation](https://github.com/data-8/datascience/issues/163) as it adds documentation for Table.set_format.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -514,7 +514,195 @@ class Table(collections.abc.MutableMapping):
     ############
 
     def set_format(self, column_or_columns, formatter):
-        """Set the format of a column."""
+        """
+        Set the pretty print format of a column(s) and/or convert its values.
+
+        Args:
+            ``column_or_columns``: values to group (column label or index, or array)
+
+            ``formatter``: a function applied to a single value within the
+                ``column_or_columns`` at a time or a formatter class, or formatter
+                class instance.
+
+        Returns:
+            A Table with ``formatter`` applied to each ``column_or_columns``.
+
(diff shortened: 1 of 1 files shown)


datascience/tables.py
```diff
@@ -3052,7 +3219,6 @@ def draw_hist(values_dict):
                         #There's a grouping in place but we're only plotting one column's values
                         label_not_grouped = [l for l in self.labels if l != group][0]
                         axis.set_xlabel(label_not_grouped + x_unit, fontsize=16)
-                    else:
```
adnanhemani, line 3055: Typo?

datascience/tables.py
```diff
@@ -514,7 +514,174 @@ def last(self, label):
     ############
 
     def set_format(self, column_or_columns, formatter):
-        """Set the format of a column."""
+        """
```
adnanhemani, line 517: Could you please add an example of one of the built-in formatter classes near the top of this docstring? Most of the users of our library, I image, would be using these quite frequently.

adnanhemani (changes requested):
Thanks for this! Just a few comments

hmstepanek:
I've made the first example use the builtin CurrencyFormatter and also made the custom formatter examples more relatable since the builtin CurrencyFormatter would not be able to handle these cases (since it always prepends the symbol). Also fixed the typo.

adnanhemani (approved):
Nice work - thanks for making this documentation so detailed!

PR #460, opened 2020-09-26
Added some more documentation in Table which was missing.

Prad06:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/docs/tables.rst
+++ b/docs/tables.rst
@@ -75,7 +75,7 @@ Accessing values
     Table.labels
     Table.first
     Table.last
-
+    Table.values
     Table.column_index
     Table.apply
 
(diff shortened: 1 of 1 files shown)


hmstepanek:
Note the following are also missing-we might want to also fix those at the same time:
* Table.column_labels
* Table.values
* Table.with_relabeling: This is deprecated and doesn't have a doc string so I'm not sure if this one is worth it.

adnanhemani:
I agree with @hmstepanek - let's make all changes in one shot via this PR. Let's hold off on `Table#with_relabeling` since it is deprecated. Good call :)

Issue #459, opened 2020-09-24
.sort(, descending=True) is not a stable sort


davidwagner:
`.sort('X')` uses a stable sort: in the case of ties in column X, it preserves the existing order of the rows.

`.sort('X', descending=True)` is not a stable sort: in case of ties, it reverses the order of the rows.  It appears to be behave as though we did `.sort('X')` and then reversed the order of the rows.

This is a bit surprising and counter-intuitive.  Is it intentional?  Is there an argument for why this behavior is better than the alternative of making `descending=True` also use a stable sort?

papajohn:
Seems like a bug to me. (Probably one that I introduced.)

hmstepanek:
I can take a closer look at this.

PR #461, opened 2020-09-26
Fix sort descending bug

hmstepanek:
- [x] Add a test that exposes the [sort descending bug](https://github.com/data-8/datascience/issues/459).
- [x] Fix the bug by reversing the order of the array before sorting and inverting the indices after.
- [x] Bump version number to 0.16.2.

This is a fix for https://github.com/data-8/datascience/issues/459.

I have explored several other options to fix this:

1. Negate the array
    ```
    if descending:
            column = -column
    ```
    The issue with this approach is that it only works on numeric columns, you cannot negate letters.
1. Use sortargs order argument and a structured array to specify how ties are handled.
    ```
    if descending:
            reverse_indexed_column = [(v, len(column)-i) for i, v in enumerate(column)]
            dt=[('value', column.dtype.str), ('index', '<i8')]
            structured_array = np.array(reverse_indexed_column, dtype=dt)
            row_numbers = np.argsort(structured_array, axis=0, kind='mergesort', order=["value", "index"])
            row_numbers = np.array(row_numbers[::-1])
    ```
    This approach works but it is quite a lot more code than the third approach taken in this PR and also requires making new arrays which is a performance hit for large data sets.


adnanhemani (approved):
Really nice and clean code - this is great! Thank you!!

PR #464, opened 2020-10-01
add documentation for append and remove functions

liudj2008:
Added documentation for table append and remove functions


adnanhemani (approved):
Looks great! Thanks for the help!!

PR #465, opened 2020-10-02
Issue #163: Included documentation for the to_array method with examples

krisstee:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>
Expanded documentation of the to_array() method in tables.py
It goes over what is returned and any arguments taken (in this case none). Examples are included to show the result of calling the method.

Additionally, trailing spaces have been cleaned up.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -784,8 +784,44 @@ class Table(collections.abc.MutableMapping):
         return self
 
     def append(self, row_or_table):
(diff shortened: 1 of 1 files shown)


adnanhemani (changes requested):
Hi! This documentation is great but can you please revert all non-documentation changes? (There's a bunch of changes relating to tabs and spaces, can we please not have those?) Thanks!

PR #466, opened 2020-10-02
Issue #163: Included documentation for to_array method.

krisstee:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

Changes include documentation added for the to_array method. Closed previous PR to stay consistent with the data-8/datascience branch and remove changes to whitespaces.
This PR should only include additions to the to_array method.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -2657,9 +2657,7 @@ class Table(collections.abc.MutableMapping):
     def to_array(self):
         """Convert the table to a structured NumPy array.
 
-        The elements in the resulting array preserve row and column replacement
-        in the table. The array's dtype attribute reflects both column names and
-        datatype.
+        The resulting array contains a sequence of rows from the table.
 
         Args:
             None
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -2665,7 +2665,7 @@ class Table(collections.abc.MutableMapping):
             None
 
         Returns:
-            arr: a NumPy array of shape [4,]
+            arr: a NumPy array
 
         The following is an example of calling to_array()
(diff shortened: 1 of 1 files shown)


krisstee:
I originally specified dimensions in what was returned in the method, but removed it as it's dependent on the table

datascience/tables.py
```diff
@@ -2655,7 +2655,42 @@ def to_csv(self, filename):
         self.to_df().to_csv(filename, index=False)
 
     def to_array(self):
-        """Convert the table to a structured NumPy array."""
+        """Convert the table to a structured NumPy array.
+
+        The elements in the resulting array preserve row and column replacement
```
davidwagner, line 2660: I did not  understand what "row and column replacement" means.  Actually, I'm having a hard time understanding this entire paragraph.  Can we reword?  Maybe the result is an array of rows, where each row is represented as a sequence of fields, and each field corresponds to a column label?  Or something?

datascience/tables.py
```diff
@@ -2655,7 +2655,42 @@ def to_csv(self, filename):
...
+        in the table. The array's dtype attribute reflects both column names and
+        datatype.
+
+        Args:
+            None
+
+        Returns:
+            arr: a NumPy array
+
+        The following is an example of calling to_array()
```
davidwagner, line 2670: Good example!

datascience/tables.py
krisstee: I meant to just type "row and column placement", but I just pushed a commit which the description is much simpler.

adnanhemani (approved):
LGTM! Thanks!

PR #468, opened 2020-10-03
#460 Fixed the documentation for Table.

Prad06:
[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/docs/tables.rst
+++ b/docs/tables.rst
@@ -43,7 +43,6 @@ Creation
 
     Table.__init__
     Table.empty
-    Table.from_rows
     Table.from_records
     Table.from_columns_dict
     Table.read_table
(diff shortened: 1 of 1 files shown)


docs/tables.rst
```diff
@@ -43,6 +43,7 @@ Creation
 
     Table.__init__
     Table.empty
+    Table.from_rows
```
davidwagner, line 46: .from_rows is deprecated, I don't think we should add it to the docs

davidwagner:
Good catch, good idea!  I had one small comment but if we address that, it looks good to me.  Thank you.

docs/tables.rst
Prad06: I will make the necessary changes! Thank you

adnanhemani (approved):
LGTM - thank you!

PR #472, opened 2020-10-04
Updated documentation for two remaining transformation methods

CodTan:
[ ] Wrote test for feature

**Changes proposed:**
Updated documentation and added examples within docstrings of `pivot_bin()` and `stack()` transformation methods.
For `pivot_bin()`, passing the argument `normed=True` responds with a Table and the following warning:
```
.../datascience/tables.py:1156: VisibleDeprecationWarning: Passing `normed=True` on non-uniform bins has always been broken, and computes neither the probability density function nor the probability mass function. The result is only correct if the bins are uniform, when density=True will produce the same result anyway. The argument will be removed in a future version of numpy.
  _, rbins = np.histogram(self[value_column],**vargs)
.../datascience/tables.py:1163: VisibleDeprecationWarning: Passing `normed=True` on non-uniform bins has always been broken, and computes neither the probability density function nor the probability mass function. The result is only correct if the bins are uniform, when density=True will produce the same result anyway. The argument will be removed in a future version of numpy.
  counts,_ = np.histogram(col_vals,**vargs)
```
Passing the argument `density=True` along with `normed=True` however results in a Table with no warnings.

Due to this, examples specific to `normed=True` for `pivot_bin()` method have not been included in this PR.


CodTan:
Need help! Not sure why the build is failing! :/

SambhaviPD:
@CodTan , Not sure whether it helps, but noticed the following in your travis build,

SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1123)
During handling of the above exception, another exception occurred:
URLError                                  Traceback (most recent call last)

CodTan:
@SambhaviPD Yes, I did check that. Since the issues were with tests/Maps.ipynb and test_read_table, I couldn't understand what to do with them.
Thanks anyway!

adnanhemani:
I've refreshed the cache and the tests shouldn't be failing anymore. Let's give it another try - please push a dummy change to tell Travis to rebuild this branch :)

CodTan:
Thanks a ton @adnanhemani. The build is successful now. :)

datascience/tables.py
```diff
@@ -1790,6 +1790,84 @@ def pivot_bin(self, pivot_columns, value_column, bins=None, **vargs) :
...
+            populated with 0s and 1s with respect to values from ``value_column`` 
+            distributed into specified ``bins`` and ``range``.
+            
+        Examples:
+        
+        >>> t = Table.from_records([
+	...   {
+	...     'column1':'data1',
+	...     'column2':86,
+        ...     'column3':'b',
```
adnanhemani, line 1805: Can we please get these all lined up on the same indent level?

datascience/tables.py
```diff
@@ -1812,6 +1890,72 @@ def pivot_bin(self, pivot_columns, value_column, bins=None, **vargs) :
...
+            A table whose first column consists of stacked values from column passed in
+            ``key``. The second column of this returned table consists of the column names
+            passed in ``labels``, whereas the final column consists of the data values
+            corresponding to the respective values in the first and second columns of the
+            new table.
+            
+        Examples:
+        
+        >>> t = Table.from_records([
+	...   {
```
adnanhemani, line 1912: Same as above - please fix the indentation levels :)

adnanhemani (changes requested):
This looks great! 🎉 Just 2 small cosmetic changes required and then we should be good to go!

CodTan:
Fixed the issues as suggested! :)

davidwagner:
The Travis build is failing, complaining about tables.py:1801, the doctests under pivot_bin(): `SyntaxError: closing parenthesis '}' does not match opening parenthesis '('`.  I confess I don't understand the error message.

CodTan:
@davidwagner, fixed the error. The build is successful now.

adnanhemani (approved):
Look great! Thanks so much! :D

PR #474, opened 2020-10-06
Added docs for the method as_text()

vrii14:
[ ] Wrote test for feature

**Changes proposed:**
Added documentation for the method as_text() with some examples as well

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -2574,7 +2574,7 @@ class Table(collections.abc.MutableMapping):
                 >>> table_astext = table.as_text()
 
                 >>> table_astext
-                'name | age  | height\nabc  | 12   | 5.5\nxyz  | 14   | 6\nuvw  | 20   | 5.9'
+                'name | age  | height\\nabc  | 12   | 5.5\\nxyz  | 14   | 6\\nuvw  | 20   | 5.9'
 
                 >>> type(table)
                 <class 'datascience.tables.Table'>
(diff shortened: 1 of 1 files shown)


vrii14:
Can't understand why are the builds failing.

adnanhemani:
I've refreshed the cache and the tests shouldn't be failing anymore. Let's give it another try - please push a dummy change to tell Travis to rebuild this branch :)

davidwagner:
The Travis build is failing with the error message: `ValueError: line 27 of the docstring for datascience.tables.Table.as_text has inconsistent leading whitespace: 'abc  | 12   | 5.5'`.  Can you look into that?  You might test your code with `make test` before submitting a pull request, to make sure all tests are passing.  Thank you!

vrii14:
I tried fixing the inconsistent leading whitespace error in the second commit itself. But it's still showing the same. 
Can you please guide to how to test code earlier? And what is a _make test_

vrii14:
Most probably the error is because of the "\n" which is in the output of the the method.
So to fix I can probably change "\n" to "\\n"(escape character with \n) but it will change the output of the example.
So should it be preferred?
Valid stackoverflow  thread [link](https://stackoverflow.com/questions/18772991/python-doctest-with-newline-characters-inconsistent-leading-whitespace-error)

adnanhemani:
`make test` is the command you can use to run your tests locally.  And yes, using the actual escape character is ok for the doctest.

adnanhemani (approved):
LGTM - thanks for this!

PR #479, opened 2020-10-08
Updated documentation for Displaying/Exporting

CodTan:
[ ] Wrote test for feature

**Changes proposed:**
Added documentation for the methods show(), as_html(), index_by() for exporting/displaying a Table.


davidwagner:
The tests are failing (see Travis CI logs above).  Please fix - I'd like to see all tests passing before I review this in detail.  Thank you.

CodTan:
The tests for this PR is now passing. However, I'm in doubt about the definition and the corresponding output of the method `show()`. Request someone review the same.

datascience/tables.py
```diff
@@ -2824,8 +2915,53 @@ def as_html(self, max_rows=0):
...
+	Args:
+	    ``columns_or_labels``: Name or label of a column of the Table,
+	    values of which are keys in the returned dict.
+	
+	Returns:
+	    A dictionary with values from the column specified in the argument
+	    ``columns_or_labels`` as keys. The corresponding data is a list of
+	    Row of values from the rest of the columns of the Table.
+	    
+	Examples:
```
adnanhemani, line 2930: On this example, could you please the change the table so that there is one repeated value in one of the columns that we give index by? That would make it much more clear to users that the rows would be aggregated under that repeated value

adnanhemani (changes requested):
The `show` method looks fine (thanks for flagging that to my attention) - but just the one comment below would be nice to have. Thank you for this!

adnanhemani (approved):
Made all corrections myself.

PR #482, opened 2020-10-15
Tox

gurupratap-matharu:
### Testing in Multiple Environments

Tox is an application that automates testing in multiple environments.

### Installation

Tox is available on PyPI via pip

`pip install tox`

### Configuration

This PR holds the tox.ini config file tests the package against major python versions
otherwise run `tox-quickstart`

### Run

Simple execute
`tox`

See video attached

- Tox will output the results of your tests against each environment. 
- The first time it runs, Tox takes a little bit of time to create the virtual environments, but once it has, the second execution will be a lot faster.
- Just make sure you have different python versions available on your machine

Run single environment

`tox -e py36`

Less verbose

`tox -q`

More verbose

`tox -v`

### DIFF
--- a/DEVELOPERS.md
+++ b/DEVELOPERS.md
@@ -36,6 +36,12 @@ Then, run the tests:
 
     make test
 
+Alternatively you can use tox. Its used for testing the package against different python versions in one go. Just make sure you have the python versions available globally or in your virtual environment
+
+Run tests
+
+`$ tox`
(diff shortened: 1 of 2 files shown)


davidwagner:
Should there be a video?  I didn't see it.

adnanhemani (approved):
This looks great - thanks!

Issue #476, opened 2020-10-07
Rise total repo testing to 97%


adnanhemani:
Due to the relatively-high activity on this repo, I'm putting down one of my larger goals in hopes that maybe people will help chip in on this!

I'm looking to rise the total repo testing percentage up to 97+% from the current 89% (I'll consider this issue closed at 97%). 

For those just starting out with this, here's a few pointers:

- Take a look at the [Coveralls report](https://coveralls.io/github/data-8/datascience?branch=master) on the master branch. You can click on individual files after expanding the repo "Tree" on this page.
- Each source file will show green or red if a test covers a line of code. Of course, less red is better - and adding tests to our test suite will help make those "red" lines of code into "green" ones :) 
- The test suite can be found [here](https://github.com/data-8/datascience/tree/master/tests). You can put new tests in either the respective ipynb file or the PyTest files. You should always try to instrument the test using the PyTest file first - but if the output of a function is a graph or something that isn't easily testable by using assert statements, you can put it in the ipynb file. 

Last thing, please try to make as many individual PRs on this as possible - only group test changes into the same PR if they're pertaining to the same function/functionality. Otherwise, create separate PRs for all new changes, please!

Thanks to all the new contributors to this repo - we appreciate your help!

MarthaDs:
Hi, I would like to take this issue. Can you please assign it to me?

davidwagner:
@MarthaDs, great, would be wonderful to have help!  I would not anticipate assigning a single individual to this issue -- rather, I imagine we'll get there through incremental progress, and that seems like something that we could easily have multiple people working on.  Can I encourage you to pick some part of the code and prepare a small Pull Request that increases coverage on that code?

athossampayo:
Did two PRs related, hope it's all good there. Separated in two PRs as asked too. Any problem let me know!

skp-github:
@adnanhemani @davidwagner  based on your comments, just wanted to let you guys know that i will start to work on this now

AdilGM:
Hi guys, is help still required? As I am just starting to contribute to open-source projects it would be great to have some assistance on starting.

anshpujara14:
Hello! I am willing to help over this issue if required. Since I am new to contributing it would be of great help for me too.

AstralRomance:
Hello. Can I help with this issue?
Is some there test strategy or msth like this?

Londala:
[...]

PR #490, opened 2020-11-08
Add tests for tables.boxplot method

slim-builder:
[ X ] Wrote test for feature

**Changes proposed:**
Add tests for tables.boxplot method for #476


adnanhemani (changes requested):
hi @slim-builder - thanks for this! I think we should put a test for boxplot in the `Charts.ipynb` instead of in `test_tables.py` as the final result of the method is a viz of sorts. Please feel free to make the changes on this branch itself and send it back for review!

PR #500, opened 2021-04-24
Add Map.Overlay tests to Maps.ipynb on add-map-overlay-tests branch

Londala:
[X] Wrote test for feature

**Changes proposed:**
Added test cases to test the Map.overlay() function in the Maps.ipynb file. The overlay() function itself seems to have some bugs, as the test cases were written as described in the function comments, but when run, exceptions are raised.


davidwagner:
If there are some bugs, let's get this fixed.  Would you be willing to submit an issue for each bug you discovered, explaining what the bug is?

davidwagner:
Note to self: I found this easier to review with `nbdiff -OD ...` (using nbdime)

davidwagner (approved):

PR #501, opened 2021-05-03
Fix Doc for show() method

Aashrutisingh:
[ ] Wrote test for feature

**Changes proposed:**
Fix Doc for show() method included a simple example with explanation and usage

### DIFF
--- a/datascience/Displaying methods.py
+++ b/datascience/Displaying methods.py
@@ -1,0 +1,23 @@
+
+
+import matplotlib.pyplot as plt
+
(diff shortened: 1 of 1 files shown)


davidwagner:
I regret I can't accept this pull request, as it doesn't meet our needs and doesn't match our style and format.  We're looking for documentation of methods that are defined in the datascience library.  `plt.show()` is a method from matplotlib, and we're not looking to document matplotlib methods in this library, as we expect people who use this library to use our methods, not the matplotlib methods.  We do our documentation via inline documentation where the method is defined, not a new standalone Python file.  Thank you for your willingness to help the project.

PR #502, opened 2021-05-27
Added test for Map.copy()

jr-42:
[ ] Wrote test for feature

**Changes proposed:**
New test for copy method of Map class
- Tests that the geojson attribute of the two Map objects are identical
- Tests that the two Map objects are distinct


jr-42:
Branch has been updated with change in PR #505 so that pipeline passes

davidwagner (approved):

davidwagner:
Nice, thank you for this!

PR #526, opened 2022-01-08
Remove function now returns array instead of list.(Issue 513 )

### LINKED ISSUE #513: Table.remove() converts columns into lists
When using `Table.remove()`, the type of the columns in a table become lists.  They should remain arrays.  See attached notebook for an example.

[TableRemoveOddity.zip](https://github.com/data-8/datascience/files/7174805/TableRemoveOddity.zip)

Sample code to reproduce:

```
>>> newTable = Table().with_columns(
        "Teams", make_array('Aces', 'Bears', 'Cats', 'Dogs', 'Eagles', 'Ferrets'),
        "Scores", make_array(93, 87, 92, 75, 96, 68)
    )
>>> newTable.remove(0)
>>> newTable.column('Scores')
[87, 92, 75, 96, 68]
>>> type(newTable.column('Scores'))
list
```

Looking at the implementation of `.remove()` (https://github.com/data-8/datascience/blob/master/datascience/tables.py#L1118), we can immediately see the problem:

```
for col in self._columns:
            self._columns[col] = [elem for i, elem in enumerate(self[col]) if i not in rows_remove]
```

It creates a list, not an array.   I believe this should be fixed to create an array.

(Seeing as `.remove()` is also side-effecting and we are trying to promote non-side-effecting methods where appropriate, we could consider deprecating `.remove()` anyway and pointing people to `.exclude()`.)

Thanks to Dan Kaiser for this bug report.

abhishict:
Fixes #513


davidwagner:
A number of tests are failing and I'm not quite sure why at the moment.

davidwagner:
Ideally, It would be nice to have a test to detect any regression of this issue.  We also need to bump the version number and document in the CHANGELOG.

adnanhemani (approved):

adnanhemani:
Approving since all I did was add a test case and update the CHANGELOG

PR #527, opened 2022-01-18
Fixed compatibility issue with collections.Iterable (Python>=3.10)

ildarakhmetov:
**Changes proposed:**
In Python 3.10, `collections.Iterable` is deprecated and replacd with `collections.abc.Iterable`. When using `datascience` with Python >=3.10, an error is thrown.

Example:

```
minard = Table.read_table('minard.csv')
minard.scatter('Longitude', 'Latitude')
```

Error: `AttributeError: module 'collections' has no attribute 'Iterable'`

I believe this little fix would solve it.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -4626,7 +4626,7 @@ class Table(collections.abc.MutableMapping):
             type(self).plots.append(axis)
         else:
             fig, axes = plt.subplots(n, 1, figsize=(width, height*n))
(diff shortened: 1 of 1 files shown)


adnanhemani (dismissed):

davidwagner:
Oh good catch!  Thank you for this fix, and sorry for my delay in responding.  This appears to be fixed in #534 .  Thank you again for your contribution.

Issue #514, opened 2021-09-17
hist.(group=...) triggers VisibleDeprecationWarning


davidwagner:
See e.g. fa21 lec10.

```
datascience/tables.py:920: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
  values = np.array(tuple(values))
```

TODO: Need to create a minimal reproducible example.

Maybe related to #458.

langdon:
assuming you have access to the galton.csv from data8 (could probably also be a bit more minimal)

```
from datascience import *
import numpy as np

%matplotlib inline
import matplotlib.pyplot as plots
plots.style.use('fivethirtyeight')

galton = Table.read_table('galton.csv')

def predict(h):
    nearby = galton.where('midparentHeight', are.between(h - 1/2, h + 1/2))
    return nearby.column('childHeight').mean()

def difference(x, y):
    return x - y

predicted_heights = galton.apply(predict, 'midparentHeight')
galton = galton.with_column('predictedHeight', predicted_heights)
pred_errs = galton.apply(difference, 'predictedHeight', 'childHeight')
galton = galton.with_column('errors',pred_errs)
galton.hist('errors', group='gender')
```

adnanhemani:
On further inspection, this is definitely caused by the same root cause as #458. I'm working on the issue now but I'm closing this ticket and transferring the comments over to track on that ticket alone.

PR #535, opened 2022-06-05
Issue 458: Remove deprecation warning

### LINKED ISSUE #458: .group() with a list raises deprecation warning
If you call tbl.group(['a', 'b']), you get the following warning:

```
//lib/python3.7/site-packages/datascience/tables.py:630: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  values = np.array(tuple(values))
```

This is unsightly.  Is there anything we should/can do in the datascience library to eliminate this?

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
This PR removes the deprecation warnings described in #458 and #514. 

Closes #458.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.17.3
+* Remove deprecation warning
+  - [Issue 458](https://github.com/data-8/datascience/issues/458) and [Issue 514](https://github.com/data-8/datascience/issues/514)
+
 ### v0.17.2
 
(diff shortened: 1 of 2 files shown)


davidwagner (approved):
Looks good to me.  I ran tests locally and this passes all tests.

davidwagner:
@adnanhemani I merged this, but then discovered that it broke a bunch of our tests, so I've reverted it.  I thought I checked that all tests were passing before merging it, but it seems I did something wrong.  Sorry about that.  So it looks like we need to go back to the drawing board on this.

Issue #512, opened 2021-09-07
Circle.map_table not using `colors` kwarg


willfurtado:
Running the following code will output an interactive map with one circle for each row (data point):

```
import pandas as pd
from datascience import *

# Load in sample data with labels, colors, and areas column
d = {
        "latitude": [-12.5, 41., -34., 40.25, -25.  ],
        "longitude": [ 18.5,  20.,  -64.,   45.,  135.],
        "labels": ['Angola', 'Albania', 'Argentina', 'Armenia', 'Australia'],
        "colors": ['blue', 'black', 'green', 'black', 'red'],
        "areas": [920.3094, 1.3349,  58.1243,   5.1861,  30.1415],
}

#Read data from dict into Pandas DataFrame and convert to `Table`
tbl = Table.from_df(pd.DataFrame(d))

#Plot a Circle Map using the `map_table` method
Circle.map_table(tbl)
```

While the `areas` argument is recognized, the circles themselves are not being colored appropriately; I believe I am calling the `Circle.map_table` method appropriately (per documentation). The output looks like: 

<img width="965" alt="Screen Shot 2021-09-07 at 10 45 59 AM" src="https://user-images.githubusercontent.com/56857749/132388389-1caf0afd-a47f-40a0-bbb2-4fabdd828530.png">

Clearly, the circles are not being colored according to the `colors` column from our original `Table` object.

raymondw2:
<img width="1023" alt="Screen Shot 2021-09-07 at 12 14 41 PM" src="https://user-images.githubusercontent.com/56237041/132398730-4a6d2cbf-dd4b-4989-b3d1-bf8f953edb61.png">
<img width="979" alt="Screen Shot 2021-09-07 at 12 14 53 PM" src="https://user-images.githubusercontent.com/56237041/132398758-df7ecd6d-6cff-4ae7-9280-eb8b84d7eb64.png">


The input to `Circle.map_table` is correct as you have written. It seems that when the size of the markers are small, it is hard to see the colors, but they are indeed there, I changed the size of the markers and it is more visible. The color argument is only for the area, not the border.

davidwagner:
Maybe we should extend `map_table` to copy the `colors` column over into a border argument, if there is no border argument explicitly specified, so that the color is more visible.  What do you think of that idea?

(I also wonder if maybe this behavior changed at some point in the past, e.g., in folium or something...)

adnanhemani:
[...]

PR #536, opened 2022-06-05
Issue 512: Make Circle marker line color default to the fill color

adnanhemani:
[X] Wrote test for feature

**Changes proposed:**
This will make map.Circle's default line color the fill color, if it is defined. Currently, we default to blue. Given that this feature is not well-documented, it is in the users' best interest to change the behavior to make the fill and line colors the same unless they are separately defined.


adnanhemani:
Closes #512.

adnanhemani:
Image of resulting change:

Screen Shot 2022-06-05 at 2 01 26 AM

davidwagner:
The code looks reasonable to me, but after running tests locally, I get a bunch of failing tests.  e.g.,

```
__________________________ tests/Charts.ipynb::Cell 2 __________________________
Input:
t.group('yr').plot('yr')

AttributeError: module 'collections' has no attribute 'Iterable'
```

Looks like this needs to merge in the fix from 5258d45dd.  I can't remember how to do that in git...

Also:

```
______________________ tests/plotly-charts.ipynb::Cell 35 ______________________
Input:
t.select(['weight', 'displ', 'hp', 'accel']).scatter3d('weight', 'displ')

RuntimeError: scatter3d is a wrapper for iscatter3d and can only be called when interactive plots are enabled
```

and

```
______________________ tests/plotly-charts.ipynb::Cell 53 ______________________
Input:
t.with_column('weight-1000', t['weight']-1000).hist(['weight', 'weight-1000'], unit='kg', width=1000, height=500)

ValueError: Image size of 72000x36000 pixels is too large. It must be less than 2^16 in each direction.
```

and more.

adnanhemani:
@davidwagner - I've merged the changes from `master` into this branch and the CI is now working! Please review and let me know if there's anything else we need from me :)

davidwagner (approved):

PR #552, opened 2022-08-13
Documentation added for Rows class.

Shershon25:
Have added documentation for the Rows class. Suggestions and corrections are most welcomed.


datascience/tables.py
```diff
@@ -5719,12 +5719,48 @@ def asdict(self):
...
 
     class Rows(collections.abc.Sequence):
-        """An iterable view over the rows in a table."""
+        """
+        An iterable row-styled view of the table object that is passed as an argument.
+
+        args:
+        table- accepts a table object of class Table()
+
+        >>>  t = td.Table().with_columns({
```
adnanhemani, line 5728: I'm unsure why we're not instantiating the `Table` class directly here. What's the `td` object here? I don't see it anywhere in the docs?

datascience/tables.py
```diff
@@ -5719,12 +5719,48 @@ def asdict(self):
...
+            returns:
+            Returns a Row object containing the i-th row of the given table
+
+            >>> t = td.Table().with_columns({
+                'letter': ['a', 'b', 'c', 'z'],
+                'count':  [  9,   3,   3,   1],
+                'points': [  1,   2,   2,  10],
+            })
+
+            >>> td.Table().Rows(t).__getitem__(0)
```
adnanhemani, line 5761: You can reuse objects from previous docs. Let's save one Rows object in the class initializer and reuse that for the remaining methods in this class.

datascience/tables.py
```diff
@@ -5735,9 +5771,35 @@ def __getitem__(self, i):
             return self._row(c[i] for c in self._table._columns.values())
 
         def __len__(self):
+            """
+            Returns the no.of rows in the table.
```
adnanhemani, line 5775: please no short-forms in the docs, as much as possible

datascience/tables.py
```diff
@@ -5735,9 +5771,35 @@ def __getitem__(self, i):
...
+
+            >>>  t = td.Table().with_columns({
+             'letter': ['a', 'b', 'c', 'z'],
+             'count':  [  9,   3,   3,   1],
+             'points': [  1,   2,   2,  10],
+            })
+            >>> td.Table().Rows(t).__len__()
```
adnanhemani, line 5782: We can jus use the Python `len` function to demonstrate this function's logic

datascience/tables.py
```diff
@@ -5719,12 +5719,48 @@ def asdict(self):
             return collections.OrderedDict(zip(self._table.labels, self))
 
     class Rows(collections.abc.Sequence):
-        """An iterable view over the rows in a table."""
+        """
+        An iterable row-styled view of the table object that is passed as an argument.
```
adnanhemani, line 5723: Let's keep the original statement here - the intended/standard usage of this class doesn't have users manually creating this class.

adnanhemani (changes requested):
Thanks for this! Let's try to make the revisions here and then we should be in a good state to merge!

datascience/tables.py
Shershon25: That was a mistake on my side. I didn't add the proper initialization for Table() class. I have updated it now.

datascience/tables.py
Shershon25: Noted.

Shershon25:
@adnanhemani I have updated the docs according to the changes suggested.

datascience/tables.py
```diff
@@ -5678,12 +5678,46 @@ def asdict(self):
...
 
     class Rows(collections.abc.Sequence):
-        """An iterable view over the rows in a table."""
+        """
+        An iterable view over the rows in a table.
+
+        args:
+        table- accepts a table object of class Table()
+        
+        >>> from datascience import Table
```
adnanhemani, line 5687: `Table` should already be imported from other doctests above, no need to re-import it.

datascience/tables.py
```diff
@@ -5678,12 +5678,46 @@ def asdict(self):
...
+        >>> t = Table().with_columns({
+             'letter': ['a', 'b', 'c', 'z'],
+             'count':  [  9,   3,   3,   1],
+             'points': [  1,   2,   2,  10],
+        })
+
+        >>> r = Table.Rows(t)
+            print(r)
```
adnanhemani, line 5695: In doctests, we don't need to `print` any values. You can simply call `>>> r` and it will output the right representation of the object.

datascience/tables.py
```diff
@@ -5678,12 +5678,46 @@ def asdict(self):
             return collections.OrderedDict(zip(self._table.labels, self))
 
     class Rows(collections.abc.Sequence):
-        """An iterable view over the rows in a table."""
+        """
+        An iterable view over the rows in a table.
+
+        args:
+        table- accepts a table object of class Table()
```
adnanhemani, line 5685: Can we please follow the standard formatting for args/return values? Here's a good example: https://github.com/data-8/datascience/blob/master/datascience/tables.py#L380-L402

E.g.:
```
...

Args:
    table (str): ...
```

adnanhemani: Also "Table" here can be written as Table or `Table` - but no parenthesis please.

datascience/tables.py
```diff
@@ -5678,12 +5678,46 @@ def asdict(self):
...
+        """
         def __init__(self, table):
             self._table = table
             self._labels = None
 
         def __getitem__(self, i):
+            """
+            Access the i-th row of the given table.
+
+            args:
```
adnanhemani, line 5711: Format comment, same as above

datascience/tables.py
```diff
@@ -5678,12 +5678,46 @@ def asdict(self):
...
+            i- index of the Row that needs to be accessed.
+
+            returns:
+            Returns a Row object containing the i-th row of the given table
+
+            >>> r.__getitem__(0)
```
adnanhemani, line 5717: To show `__getitem__`, you can just use the `[]` notation. Example: `r[0]` should give you the same result. Change required because it is more Pythonic and easier to read/remember for beginning learners.

datascience/tables.py
```diff
@@ -5694,9 +5728,24 @@ def __getitem__(self, i):
...
+
+            """
             return self._table.num_rows
 
         def __repr__(self):
+            """
+            Returns the printable representation of the given table as string.
+            Uses the standard repr() function.
+
+            >>> r.__repr__()
```
adnanhemani, line 5745: Similar as the last comment, but we should show this as `repr(r)` instead.

adnanhemani (changes requested):
Hi, getting closer! In addition to the comments, the doc build is now failing as well :/ (check the GH Action attached to the repo). Please fix these, thank you!

Shershon25:
Hey! I have updated the docs again. I corrected the incorrect prefix because of which the build was failing. If any other corrections are there, do let me know.

adnanhemani:
Hi! I think this is awesome!! Style and comments-wise, I think I'm ready to approve. But the build is failing (see above). Can you please take a look and ensure that all tests pass? Once that's done, I think I can approve and merge! Thank you for your hard work on this!

adnanhemani:
Hi, your changes were still failing the build, so I went ahead and made the required changes to get it to work. Thank you for this work!!

adnanhemani (approved):

PR #553, opened 2022-09-08
Add pythonic aliases for comparison predicates.

EricApostal:

**Changes proposed:**
Add pythonic aliases for comparison predicates.
Ex: added `are.greater_than` as an alias of `are.above`


davidwagner:
Can you help me understand the rationale for why this addition is needed/beneficial?

SirTZN:
I would also like to add that this is the notation already used in the some of the function's descriptions.
Snippet from predicates.py:

```python
@staticmethod
def above(y):
    """Greater than y."""
```

adnanhemani (changes requested):
I can agree to seeing that this might be beneficial and it's great that it's just adding a simple alias, so there's no code maintainability issues. I just want to ensure that code testing coverage isn't hit as a result of this new code, so can we please add some testing code around the new methods?

Once we have the testing code, I can merge this in, it's in great condition. Thanks!

SirTZN:
I added methods replicating the existing ones. I've never worked with pytest, however, so please do let me know if anything needs to be amended.

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.17.7
+* Add pythonic aliases for comparison predicates.
+* Add alias tests to test_predicates.py
```
adnanhemani, line 9: We don't need this line. Just the first one is fine :)

datascience/predicates.py
```diff
@@ -211,6 +211,49 @@ def not_containing(substring):
...
+        """Greater than y."""
+        return are.above(y)
+
+    @staticmethod
+    def greater_than_or_equal_to(y):
+        return are.above_or_equal_to(y)
+
+    @staticmethod
+    def less_than_or_equal_to(y):
+        return are.below_or_equal_to(y)
```
adnanhemani, line 235: Based on: https://coveralls.io/builds/55559799/source?filename=datascience%2Fpredicates.py

We're missing test coverage on this method and a few others. The tests you've created are fantastic so far, if you could please add a few more for the missing test methods - this should be mergeable!

adnanhemani (changes requested):
Hey, this is very close to being mergable - thank you! You should be able to see updated coverage when you push new commits now.

adnanhemani:
Closing due to inactivity.

PR #555, opened 2022-09-21
this package actually requires atleast 3.6 because fstrings are used

zahash:
[ x ] Wrote test for feature
[ x ] Added changes to CHANGELOG.md

**Changes proposed:**
this package actually requires atleast 3.6 because fstrings are used
also, added python_requires attribute in setup (issue https://github.com/data-8/datascience/issues/529)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,8 @@ from setuptools import setup
 from setuptools.command.test import test as TestCommand
 
 
-if sys.version_info < (3, 0):
-    raise ValueError('This package requires python >= 3.0')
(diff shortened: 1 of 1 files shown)


davidwagner:
@adnanhemani What do you think about setting a requirement for Python 3.6?

adnanhemani (changes requested):
Hio @zahash - thanks for this. As per #559, can we please increment the package version and add the change to the CHANGELOG? Thanks!

davidwagner:
Fixed by #564, and no longer needed.  Thank you!

PR #558, opened 2022-10-02
Remove Deprecated Function for AttributeError

ajgrover:
[ ] Wrote test for feature
[ x ] Added changes to CHANGELOG.md

**Changes proposed:**
Replaced deprecation warning when calling non-Table attributes on Table with an AttributeError. If you currently call .item(x) on a Table, it will call .item(x) on each column thus outputting Table with row x. This functionality is undocumented and breaks the abstraction barrier.


davidwagner:
Seems like a reasonable goal.  Why not simply remove the code of `Table.__getattr__`?  Would that accomplish the same goal?

ajgrover:
> Seems like a reasonable goal. Why not simply remove the code of `Table.__getattr__`? Would that accomplish the same goal?

I think it would. My reasoning for retaining it was to add a more detailed Attribute Error in case someone had been using it unknowingly. In that case, I imagine there is a deeper conceptual understanding issue that an error message wouldn't be able to fix. I've removed it entirely now and adjusted the changelog accordingly.

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.17.6
+* Removed deprecated function allowing calling non-Table attributes on a Table. Will now throw an AttributeError.
```
adnanhemani, line 8: Can we please remove the second sentence here since we've changed the functionality? Other than that, looks great. Thanks!

adnanhemani (changes requested):

CHANGELOG.md
ajgrover: should be good now!

adnanhemani:
Looks great, thank you!

adnanhemani (approved):
LGTM

PR #563, opened 2022-11-21
Added test for util.py is_non_string_iterable() function #476

Lahiry:
Added test case for `util.py` is_non_string_iterable() function in `test_util.py`


adnanhemani (changes requested):
Hi, this is really great! But can you please one additional test case for a `collections.abc.Sequence` type? The reason why I'm requiring it is because I want full coverage on this method as per Coveralls (https://coveralls.io/builds/50171925/source?filename=datascience%2Futil.py#L257). We already have shadow coverage on this method from other tests, but if we can fully cover all branches using this additional test - that would be amazing!

Issue #559, opened 2022-10-08
We use Python f-strings


davidwagner:
The code uses f-strings (tables.py:3161, maps.py:129, and some test code), which requires Python 3.6, but we only check for Python 3.0.

Should we update the checks to require Python 3.6, or should we remove the f-strings so this works with Python 3.0?

davidwagner:
@adnanhemani I'd be interested to hear your thoughts on this.  (See also #555.)

adnanhemani:
I think we should go ahead and update to use Python 3.6 - most of the library was developed and first deployed using Python 3.6+.

Here's my recommendation - let's update the library version by a medium version (up to 0.18.0, I believe) and then make Python 3.6.0 the minimum required version. I'll make a comment on the PR for this.

davidwagner:
Done - see #564

PR #564, opened 2022-11-26
Python3 6 version

Paulokim1:
[ ] Wrote test for feature

**Changes proposed:**
Modified setup.py due to 3.0 limitations. Also, a simpler approach to use versions with use of import, taking out the need of using the with open commad and iterate through a file.


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### v0.17.6
```
adnanhemani, line 7: Hi @Paulokim1 - thanks for this. As per https://github.com/data-8/datascience/issues/559, can we please increment the package version to 0.18.0? Thanks!

adnanhemani (changes requested):

setup.py
```diff
@@ -46,7 +40,8 @@ def run_tests(self):
 setup(
     name = 'datascience',
     packages = ['datascience'],
-    version = version,
+    python_requires='>=3.6',
+    version = __version__,
```
davidwagner, line 44: Why change `version = version` to `version = __version__`?  This seems like it changes the logic of this code.  What is the purpose of this?

davidwagner (changes requested):
We need to update datascience/version.py to set the version to 0.18.0, to match the updated CHANGELOG

adnanhemani (approved):
@davidwagner - just fixed this CR up. Let me know if this works :)

setup.py
adnanhemani: This makes the build fail due to a circular dependency. I've reverted this logic now.

davidwagner (approved):

PR #565, opened 2022-11-28
Improved test_formats.py tests

ChrisFelt:
[x] Wrote test for feature

**Changes proposed:**
Improved several test_formats.py tests with random testing and edge case testing.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -3,6 +3,8 @@ import re
 import datascience as ds
 from datascience import formats
 import random
+import datetime
+import time
 
 
 def assert_equal(string1, string2):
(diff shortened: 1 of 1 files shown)


adnanhemani (changes requested):
Hi, thanks for this. We currently do not do any random testing on this repo as a result of wanting to keep the test cases as clear as possible to developers and reduce the chances of false negatives while testing. (All actions in this library should be idempotent and stateless by design, so random testing is not essential to our testing methodology either).

I do, however, like your edge case testing in other parts of the CR! If you'd be willing to remove the random testing and just keep the edge case testing, I'd be glad to merge those changes in!

PR #562, opened 2022-11-10
New tests for formats, predicates, tables and utils

RodrigoGCoelhoo:
**Changes proposed:**
New tests for the features already developed. Formats and Predicates with 100% coverage.

PR #567, opened 2022-12-07
Add coverage tests for formats.py

dengpris:
[x] Wrote test for feature

**Changes proposed:**
Added coverage tests to formats.py for issue #476

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -17,6 +17,11 @@ def test_doctests():
 
 
 def test_default_format():
+    fmt = ds.Formatter(min_width=1, max_width=2, etc=3)
+    assert_equal(fmt.min_width,1)
+    assert_equal(fmt.max_width, 2)
+    assert_equal(fmt.etc, 3)
+
     fmt = ds.default_formatter.format_value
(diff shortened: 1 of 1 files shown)


tests/test_formats.py
```diff
@@ -82,6 +87,8 @@ def test_date_format():
     t.set_format('time', ds.DateFormatter("%Y-%m-%d %H:%M:%S.%f"))
     assert isinstance(t['time'][0], float)
 
+    fmt = ds.DateFormatter("%Y-%m-%d").format_value
+    assert_equal(fmt(0), "1970-01-01")
```
adnanhemani, line 91: This yields me "1969-12-31" instead of "1970-01-01" due to time zones. Can we make the test so that it accepts either value instead? We want to make sure that tests don't fail on local machines

adnanhemani (changes requested):
Extremely clean code and you hit the right coverage branches, so bravo for that! But one change required - thanks!

adnanhemani:
Closing this in favor of #562. Thanks for the changes though! If you feel there are any edge cases here that are not covered by that PR, please highlight them and re-open this PR.

PR #568, opened 2022-12-07
Add coverage tests for tables.py

canlasla:
[x] Wrote test for feature

**Changes proposed:**
Added coverage tests to tables.py for issue https://github.com/data-8/datascience/issues/476

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1157,10 +1157,6 @@ def test_remove_column_type(table):
 # Create #
 ##########
 
-def test_empty_labels_none():
-    with pytest.raises(ValueError):
-        t = Table().empty(['letter', 'count', 'points'])
-
 def test_empty():
     t = Table(['letter', 'count', 'points'])
     assert_equal(t, """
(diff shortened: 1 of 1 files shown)


tests/test_tables.py
```diff
@@ -1157,6 +1157,9 @@ def test_remove_column_type(table):
 # Create #
 ##########
 
+def test_empty_labels_none():
```
adnanhemani, line 1160: I think you've actually stumbled on a bug here - this shouldn't have thrown an Error. Can you remove this test case and I'll create a ticket in the backlog to remove this method altogether.

adnanhemani (changes requested):
Thanks for this!

tests/test_tables.py
adnanhemani: Update: I've solved this bug in #570

adnanhemani:
Closing this review as none of the test cases added are applicable to the code as it stands now. Thank you for your efforts :)

Issue #551, opened 2022-08-11
Getting warning from numpy when creating histogram


sean-morris:
I am using datascience 0.17.5 with matplotlib 3.5.2

![image](https://user-images.githubusercontent.com/1514995/184213943-5e34c656-14e9-4b39-85a4-27cad4df6757.png)

This seems to talk about a workaround:
https://stackoverflow.com/questions/63723514/userwarning-fixedformatter-should-only-be-used-together-with-fixedlocator

SakshiSingh25:
Hi, I am interested in working on this issue. Kindly, guide me in the further process.
Thank you.

ToshiNandanReddy:
Hello, I am interested in working on this issue.  Could you please guide me with further steps.
Thank you.

helloaadii:
Hello, I am interested in working on this issue. Could you please guide me with further steps.
Thank you.

davidwagner:
I'm afraid we don't have resources to guide you on working on this.  Sorry.

adnanhemani:
Can confirm that I can reproduce this on datascience 0.17.6 and matplotlib 3.5.3. Minimalistic example code:
```
from datascience import *
import matplotlib.pyplot as plt
%matplotlib inline
plt.style.use('fivethirtyeight')
from bokeh.sampledata.autompg import autompg as df
t = Table.from_df(df)
t["mpg_re"] = t["mpg"] * 100000000
t.hist('mpg_re')
```

Investigating fix options now.

adnanhemani:
Closed by #574. Thanks for the bug report @sean-morris!

PR #574, opened 2022-12-25
Fixes UserWarning for certain charts

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
This fixes the warning described in #551. This is caused due to a new change in the matplotlib library that shows this UserWarning when not explicitly supplying the ticks with the tick labels. Although this does not cause any detrimental action to the user, we should remove such user warnings when possible.

In this PR, I have make it so that we always set the ticks before the tick labels to the plot. I have tested the change manually in the test/Charts.ipynb as well to ensure that all chart types affected by this UserWarning are fixed. Here's the old chart versus the new chart:

### Old
Screen Shot 2022-12-25 at 12 02 17 PM

### New
Screen Shot 2022-12-25 at 12 02 48 PM

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -3620,7 +3620,8 @@ class Table(collections.abc.MutableMapping):
             axis.bar(index-0.5, self[label], 1.0, color=color, **options)
 
         def annotate(axis, ticks):
-            if (ticks is not None) :
+            if (ticks is not None):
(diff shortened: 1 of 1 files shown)


adnanhemani:
Closes #551.

davidwagner (approved):

PR #576, opened 2023-01-10
testing coverage and cleanup updates tables.py

amoghkokari:
removing depreciated columns_lables. It is not being used anywhere in the code. Should help in boosting the code coverage.

refactoring the following code to avoid redundancy 

`if not hasattr(formatter, 'format_column'):`
`    if callable(formatter):`
`         formatter = _formats.FunctionFormatter(formatter)`
`     raise Exception('Expected Formatter or function: ' + str(formatter))`

[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -746,9 +746,9 @@ class Table(collections.abc.MutableMapping):
         """
         if inspect.isclass(formatter):
             formatter = formatter()
+        if not hasattr(formatter, 'format_column') and callable(formatter):
+            formatter = _formats.FunctionFormatter(formatter)
         if not hasattr(formatter, 'format_column'):
-            if callable(formatter):
-                formatter = _formats.FunctionFormatter(formatter)
             raise Exception('Expected Formatter or function: ' + str(formatter))
         for label in self._as_labels(column_or_columns):
             if formatter.converts_values:
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -344,13 +344,6 @@ class Table(collections.abc.MutableMapping):
         """
         return tuple(self._columns.keys())
 
-    # Deprecated
-    @property
-    def column_labels(self):
-        """Return a tuple of column labels. [Deprecated]"""
-        warnings.warn("column_labels is deprecated; use labels", FutureWarning)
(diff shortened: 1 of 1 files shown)


datascience/tables.py
```diff
@@ -753,9 +746,9 @@ class instance. The formatter is run each time ``__repr__`` is called. So, while
         """
         if inspect.isclass(formatter):
             formatter = formatter()
-        if callable(formatter) and not hasattr(formatter, 'format_column'):
-            formatter = _formats.FunctionFormatter(formatter)
         if not hasattr(formatter, 'format_column'):
+	    if callable(formatter):
```
adnanhemani, line 750: This is incorrect - the `formatter` variable was being overwritten in the current version of the code. If the overwritten still didn't have the `format_column` attr, then you'd want to raise the exception. Please revert this change - or explain further why we should introduce new logic to this function.

In either case, the repo is not compiling due to the changes here (look at the GH Action results) - so please make a change here.

adnanhemani (changes requested):
Thanks for this and your interest in helping with this repo! Please make the required changes.

datascience/tables.py
amoghkokari: Got it, yes it is being overwritten, reverted the change

datascience/tables.py
```diff
@@ -753,7 +746,7 @@ class instance. The formatter is run each time ``__repr__`` is called. So, while
         """
         if inspect.isclass(formatter):
             formatter = formatter()
-        if callable(formatter) and not hasattr(formatter, 'format_column'):
+        if not hasattr(formatter, 'format_column') and callable(formatter):
```
adnanhemani, line 749: Can we revert this properly to ensure that the order of the checks is the same as before? This could cause issues that would've been averted due to Python short-circuiting.

adnanhemani (changes requested):

datascience/tables.py
amoghkokari: Sorry for the delay, yes, I have reverted this change.

adnanhemani (approved):
LGTM

PR #578, opened 2023-02-09
Feature/additional tests

FlynnOwen:
[x] Wrote test for feature

**Changes proposed:**
I've added additional unit tests to increase coverage.
FunctionFormatter and is_non_string_iterable had no tests, and I added additional tests for tautologies and contradictions to for the predicates module.

### DIFF
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1,4 +1,5 @@
 import doctest
+import pytest
 import re
 import datascience as ds
 from datascience import formats
@@ -106,3 +107,11 @@ def test_distribution_formatter():
     30.00%
     38.33%
     """)
+
+
+def test_function_formatter():
+    def _mult_ten(v):
+        return v * 10
+
+    func_formatter = formats.FunctionFormatter(_mult_ten)
+    assert func_formatter.format_value(5) == '50'
--- a/tests/test_formats.py
+++ b/tests/test_formats.py
@@ -1,5 +1,4 @@
 import doctest
-import pytest
 import re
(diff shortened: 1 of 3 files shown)


tests/test_formats.py
```diff
@@ -1,4 +1,5 @@
 import doctest
+import pytest
```
adnanhemani, line 2: Please don't add new import statements unless they're absolutely necessary

tests/test_util.py
```diff
@@ -96,3 +96,14 @@ def test_proportions_from_distribution():
...
+
+def test_is_non_string_iterable():
+    is_string = 'hello'
+    assert ds.is_non_string_iterable(is_string) == False
+
+    is_list = [1, 2, 3]
+    assert ds.is_non_string_iterable(is_list) == True
+
+    is_int = 1
+    assert ds.is_non_string_iterable(is_int) == False
```
adnanhemani, line 109: Can we add the remaining case that's not covered by tests right now: https://coveralls.io/builds/56773172/source?filename=datascience%2Futil.py#L257

adnanhemani (changes requested):
Hi, thank you for this. Can you please address these comments and we should be okay to merge!

tests/test_util.py
FlynnOwen: Thanks for the feedback @adnanhemani, I've removed the import and updated with a unit test for the Sequence type.

adnanhemani (approved):
LGTM. Thank you!

PR #582, opened 2023-02-11
Adds a test to check if an empty table is created from records if empty record list

aad1tya:
#476 
Table().from_records() should return an empty table.
[*] Wrote test for feature

**Changes proposed:**
To increase the code coverage I have added a test to check if an empty table is created when 
an empty record is passed.


adnanhemani (approved):
LGTM. Thanks!

Issue #586, opened 2023-04-17
test_date_formatter_format_value() should call tzset()


davidwagner:
The test for `test_date_formatter_format_value()` failed on my system:

```
E       AssertionError: 
E         2022-10-20 04:14:49.900400
E         !=
E         2022-10-20 11:14:49.900400
```

But when I explicitly set the TZ environment variable (`export TZ=UTC`), then run the tests, the test passes.

I think after setting `os.environ["TZ"] = "UTC"`, it is necessary to call `time.tzset()`.  See https://docs.python.org/3/library/time.html#time.tzset.

See this test to demonstrate that just setting `os.environ` isn't enough:

```
$ cat test_tz.py 
from datetime import datetime, timezone
import os

print(os.environ["TZ"] if "TZ" in os.environ else "unset")
print(datetime.fromtimestamp(1666264489.9004).strftime("%Y-%m-%d %H:%M:%S.%f"))

os.environ["TZ"] = "UTC"

print(os.environ["TZ"] if "TZ" in os.environ else "unset")
print(datetime.fromtimestamp(1666264489.9004).strftime("%Y-%m-%d %H:%M:%S.%f"))
$ echo $TZ

$ python3 test_tz.py 
unset
2022-10-20 04:14:49.900400
UTC
2022-10-20 04:14:49.900400
$ export TZ=UTC
$ echo $TZ
UTC
$ python3 test_tz.py
UTC
2022-10-20 11:14:49.900400
UTC
2022-10-20 11:14:49.900400
```

If I add a call to `time.tzset()`, then everything works as expected.

adnanhemani:
Merged #587. Thanks for this! It was working fine on my machine, so I didn't catch this myself :)

PR #587, opened 2023-04-17
Fix bug #586 - need to call tzset() after setting TZ

davidwagner:
**Changes proposed:**
Fix bug in testing code, so it works across platforms, by adding a call to tzset() after setting the TZ environment variable.


adnanhemani (approved):
LGTM

PR #589, opened 2023-06-01
Tests that feature is converted to list or not

Devasy23:
This commit adds a new test function, test_map_feature_list, to the test_maps.py file. The function tests whether the features variable in the datascience.maps module is correctly converted to a list. The test creates a Map object, adds a Marker feature to it, and then checks that the features list contains the Marker feature. This test ensures that the features variable is properly converted to a list and helps maintain code quality.

[ ] Wrote test for feature

**Changes proposed:**
<description of the PR>

### DIFF
--- a/tests/test_maps.py
+++ b/tests/test_maps.py
@@ -55,6 +55,24 @@ def test_map_marker_and_region(states):
     ds.Map([region]).show()
     ds.Map([marker, region]).show()
 
+    
+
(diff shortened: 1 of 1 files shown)


adnanhemani (changes requested):
Hi, have you tried running this test locally? I'm not seeing an `add_feature` function to `Map`. Did you mean `Map().overlay`?

PR #595, opened 2023-08-06
Add test for when None is appended to Table object

tristannew:
[ ] Wrote test for feature

**Changes proposed:**
This PR adds a simple test case to `test_tables.py` that asserts that, when a `None` type is appended to a Table, it does not return anything, and the Table remains the same.


adnanhemani (approved):
LGTM

PR #597, opened 2023-08-18
_varargs_labels_as_list test

tiddoloos:
[ ] Wrote test for _varargs_labels_as_list

**Changes proposed:**
tests for _varargs_labels_as_list

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1938,3 +1938,26 @@ def test_no_records():
     empty_table = Table()
     records_empty_table = Table().from_records([])
     assert empty_table == records_empty_table
+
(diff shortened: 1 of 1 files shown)


adnanhemani (changes requested):
In most languages, testing "private" or "internal" functions is not a pattern that is encouraged (if that pattern even allows it at all). While I appreciate the additional test coverage these tests provide greatly, I think it would be best if we made test cases using `Table#drop` and `Table#select` to trigger all the conditional statements within `Table#_varargs_labels_as_list` instead. Let me know your thoughts.

adnanhemani:
Closing PR as stale.

PR #598, opened 2023-08-18
Adding tests for predicates functions

BurakCakan:
[x] Wrote test for feature
I checked the CHANGELOG file but I couldn't see any example of writing addition of tests, so I didn't add it.

**Changes proposed:**
Tests are added for the functions of predicates.py file including check_iterable, not_containing and containing.


adnanhemani (commented):
Tests for these functions exist in the doc tests - https://github.com/data-8/datascience/blob/master/datascience/predicates.py#L17-L102

So there's no reason to add additional explicit tests here. Thanks for this anyways - please feel free to use our Coveralls to see where there might be more code that does not currently have a test - mostly Maps and Tables need the most help right now!

PR #601, opened 2023-09-22
Added test for table copy method

AlexMuresan:
[x] Wrote test for feature

Changes proposed:
Test for Table copy method

Tests that a deep copy is not changed when original table is changed Tests that a shallow copy is changed when the original table is changed


adnanhemani:
Hi, thank you for this - but we already have a test for shallow copy vs regular copy as part of the doctests in the code itself. So this addition to the test suite is not really required. If you feel differently, please let me know why this adds value. Thanks!

PR #602, opened 2023-09-23
Added tests for table sample with distribution method

AlexMuresan:
[x] Wrote test for feature

Changes proposed:
Test for Table sample_from_distribution method

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1716,6 +1716,40 @@ def test_shuffle_different_order(table):
             return
     assert False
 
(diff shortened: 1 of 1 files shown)


adnanhemani:
Hi, closing this PR - it's quite literally a copy of the doctests so doesn't add any value to the test suite. If you wish to contribute value to our test suite, please take a look at the Coveralls and find LOCs that are untested and create test cases around those. For reference, here is the Coveralls: https://coveralls.io/builds/62164828/source?filename=datascience%2Ftables.py#L2322

PR #609, opened 2024-02-12
tables.py test

add-itya:
[X] Wrote test for feature

**Changes proposed:**
I added functionality to check column entry types are consistent. I saw that there are functions that check the values of the basic tables, and thought it would be nice to add a check for their types. Please let me know if anything is missing, thanks :)

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -137,6 +137,12 @@ def test_basic(table):
     z      | 1     | 10
     """)
 
+def test_column_types(table):
+    """Tests that the column types are correct."""
+    t = table
+    for col_name in t.labels:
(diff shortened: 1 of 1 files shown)


adnanhemani:
Hi @add-itya - given that the columns are a NumPy array, it is pretty much a requirement that they have the same object types. Given that, I'm inclined to close this PR, as it is unlikely that there is any change related here that may still be useful to the repo.

Did you see any test coverage that is not covered without this check? I'm glad to take a look if that's the case - I wasn't able to verify quickly as the GitHub Action above (which runs the test suite and coverage checker) failed. Please take a look at that if you think this change still helps with the code coverage.

PR #610, opened 2024-02-15
Updating Readme.md file to detail how to get local testing/coverage report

add-itya:
[X] Updated Readme.md file

**Changes proposed:**
Before I started developing, I noticed that my Windows OS caused some issues when generating the testing reports. Using pytest/coverage, I got different percentages and even errors. The errors were because some functions that were being used were only available on Unix/Linux systems. So, I had to install WSL and work around this to gain an accurate report. This way, I could add tests more effectively and help get the overall testing coveralls report to 97%.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,3 +63,9 @@ Optionally, integrate with Coveralls for code coverage tracking:
 python3 coveralls
 
 That's it! You have now successfully set up your local development environment and generated a test report.
+
+After reviewing the results, you can determine which of the files under the source code folder need more tests based on their corresponding percentages.
+
+Our aim is to have 97% of overall codebase coverage to ensure that there are no discrepencies.
+
+Thank you for considering to help our project, we really appreciate it :)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,3 +21,51 @@ pip install datascience
...
+
+6. View coverage report:
+python3 coverage report
+
+Optionally, integrate with Coveralls for code coverage tracking:
+python3 coveralls
+
+That's it! You have now successfully set up your local development environment and generated a test report.
+
+After reviewing the results, you can determine which of the files under the source code folder need more tests based on their corresponding percentages.
```
adnanhemani, line 67: No need to add the lines from here onwards :)

adnanhemani (changes requested):
Hi @add-itya - so sorry for the delay to review this. Can you please merge these into the `DEVELOPERS.md` file instead? The `README.md` is supposed to be a much more general file and people working on the library (including testing) should ideally be looking at that file instead. Please let me know if you have any questions. Thanks for this - I'm sure it will be very helpful to Windows developers in the future!

adnanhemani:
Closing due to inactivity.

PR #611, opened 2024-02-27
Add test for tables.remove method

sinchana-kumbale:
- [x] Wrote test for feature
- [ ] Added changes to CHANGELOG.md

**Changes proposed:**
Added a test for `tables.remove` in line with issue #476 
The test ensures method returns None and modifies no rows when called with an empty row list

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1159,9 +1159,9 @@ def test_remove_single(table):
     """)
 
 def test_remove_none(table):
-	num_rows_before = table.num_rows
-	assert table.remove([]) is None
-	assert table.num_rows == num_rows_before
+    num_rows_before = table.num_rows
+    assert table.remove([]) is None
+    assert table.num_rows == num_rows_before
 
 def test_remove_zeroth_row(table):
     table.remove(0)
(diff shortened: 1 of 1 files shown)


tests/test_tables.py
```diff
@@ -1158,6 +1158,11 @@ def test_remove_single(table):
     z      | 1     | 10
     """)
 
+def test_remove_none(table):
+	num_rows_before = table.num_rows
```
adnanhemani, line 1162: Seems like your tabs setting is set to 8 spaces versus the 4 spaces in the rest of this file. Can you please recheck this and make sure that the tabs look uniform?

adnanhemani (dismissed):
This is great! Just one comment.

tests/test_tables.py
sinchana-kumbale: Thanks for the feedback on the tab spacing! I've made the changes to ensure consistency. Could you please take another look and confirm if it looks correct on your end?

adnanhemani: Unfortunately, it doesn’t look like your new commit did anything :( the diff is empty and the indent still looks like 8 spaces (check the commits tab here on the PR). Can you please try again?

sinchana-kumbale: You were right, I did check and the commit had not changed the spacing. Please verify now. Thank you

davidwagner: Looks good to me, thank you!

davidwagner (approved):

PR #613, opened 2024-05-30
added tests for branching in tables _join

rwmehta:
[X] Wrote test for feature

**Changes proposed:**
Added tests for if and if not branches in tables _join method as requested in issue #476 . 
Ensures the method returns None when either table has no rows, and matches the label when "other_label" is not provided.

### DIFF
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1631,6 +1631,36 @@ def test_move_column(table):
     table = table.move_column(2, 1)
     assert table.labels == ('count', 'points', 'letter')
 
+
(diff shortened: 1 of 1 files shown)


davidwagner:
This doesn't seem to increase test coverage.  I think these cases are already covered by other pre-existing test cases: empty tables by `test_join_no_rows()`, and no 3rd argument to join by many other test cases.  But OK, why not, more tests can't hurt.  Thanks for the contribution.

davidwagner (approved):

PR #614, opened 2024-09-26
Update README.md

Kamilo408:
Wrote new README


davidwagner:
Thanks for the suggestion. I think I'll stick with the existing README.

Issue #538, opened 2022-06-05
Make Plots Compatible with %matplotlib notebook


adnanhemani:
Currently, the plot dimensions seem to have issues when using `%matplotlib notebook`. This magic function will allow us to make all plots interactive by default. However, the plot dimensions are causing problems out-of-the-box at this time. Example:

<img width="935" alt="Screen Shot 2022-06-05 at 6 15 39 PM" src="https://user-images.githubusercontent.com/10746770/172074662-6a823049-3caa-467e-8978-17dc9d91fa1a.png">

SakshiSingh25:
Hi, I would like to work on this issue. Would you guide me in the further process? 
Thanks

Vallabh100:
That's what _**%matplotlib notebook**_ command do right? We use this command to get zoom-able and resize-able interactive plots. So even though plot dimensions do cause problem initially, we can resize the plot. If you are facing timeout problem in notebook , just by using _**%matplotlib**_, it open separate prompt of plot and there u can visualize it. See below pic-
![image](https://github.com/data-8/datascience/assets/108627108/a015467a-448b-40ea-985e-4a125b004756)

adnanhemani:
Yes, you are right that by we can resize and/or pop-out the plot - but we should fix this (if it is possible) to work with the majority of our plots right out of the box without the user needing to re-size or zoom on the plot

PR #617, opened 2024-10-04
Fix axis labels being cut off in interactive histograms

k-luka:
**Issue:**
Axis labels were being cut off in histograms when using `%matplotlib notebook`. This affected the readability of the plots.

**Fix:**
Added `plt.tight_layout()` to the `draw_hist` function in `tables.py` to ensure the layout is automatically adjusted to prevent labels from being cut off.

**Testing:**
Tested the `hist` method using both `%matplotlib notebook` and `%matplotlib inline` backends to confirm that the issue is resolved without introducing any side effects.

**Related Issues:**
Make Plots Compatible with %matplotlib notebook #538

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -5349,15 +5349,14 @@ class Table(collections.abc.MutableMapping):
         >>> t.hist('value', group='category') # doctest: +SKIP
         <two overlaid histograms of the data [1, 2, 3] and [2, 5]>
         """
+
         # Matplotlib has deprecated the normed keyword.
-        # TODO consider changing this function to use density= instead too
         if 'normed' not in vargs and 'density' not in vargs:
             vargs['density'] = True
         elif 'normed' in vargs and 'density' not in vargs:
             vargs['density'] = vargs.pop('normed')
         elif 'normed' in vargs and 'density' in vargs:
-            raise ValueError("You can't specify both normed and density. "
-                             "Use one or the other.")
+            raise ValueError("You can't specify both normed and density. Use one or the other.")
 
         global _INTERACTIVE_PLOTS
         if _INTERACTIVE_PLOTS:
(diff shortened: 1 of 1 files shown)


adnanhemani:
Hi @k-luka - thanks for this. Unfortunately, here are a lot of formatting changes (which I'd highly prefer not to have, to be honest), so it's hard to review what changes exactly you've made. Additionally, I need a screenshot proof of your testing of this (and ideally, a change in the test notebook for plots that show this case).

Given both of these, it will take a longer time than I'd like to spend to review this - so I'd like to see both of these fixed before I can dedicate time to review.

datascience/tables.py
```diff
@@ -5349,15 +5349,14 @@ def hist(self, *columns, overlay=True, bins=None, bin_column=None, unit=None, co
         >>> t.hist('value', group='category') # doctest: +SKIP
         <two overlaid histograms of the data [1, 2, 3] and [2, 5]>
         """
+
         # Matplotlib has deprecated the normed keyword.
-        # TODO consider changing this function to use density= instead too
```
davidwagner, line 5353: What is the justification for removing this TODO?

datascience/tables.py
```diff
@@ -5529,14 +5519,21 @@ def draw_hist(values_dict):
...
-                                 color=self.chart_colors[1], align="edge")
+                                color=self.chart_colors[1], align="edge")
                     _vertical_x(axis)
                     if rug:
                         axis.scatter(values_for_hist, np.zeros_like(values_for_hist), marker="|",
-                                     color="black", s=100, zorder=10)
+                                    color="black", s=100, zorder=10)
                     type(self).plots.append(axis)
 
+            # Added to ensure proaper axis label dimensions
```
davidwagner, line 5529: proaper -> proper

PR #623, opened 2025-03-10
fixed booleans from being converted to integers in make_array

### LINKED ISSUE #622: make_array with booleans will be converted to integers
Hi, current Data 8 TA here!

During a TA meeting we've realised that when creating boolean arrays with `make_array()`, the elements will be converted to integers and will not follow numpy behaviour due to the extra check for Windows machines (lines 42-45 in [`util.py`](https://github.com/data-8/datascience/blob/cf403b5582a4ed6fa1791917707bf38bf1d2de96/datascience/util.py)).

```py
if elements and all(isinstance(item, (int, np.integer)) for item in elements):
        # Specifically added for Windows machines where the default 
        # integer is int32 - see GH issue #339.
        return np.array(elements, dtype="int64")
```

You can see this behaviour as follows:
```py
In [1]: Table().with_columns("bool", make_array(True, False))
Out[1]: 
bool
1
0

In [2]: Table().with_columns("bool", np.array([True, False]))
Out[2]: 
bool
True
False
```

This is because Python behaviour will cause booleans to also return `True` for `isinstance(b, int)` and hence they will be converted to `int64` by the extra check above.
```
>>> isinstance(True, int)
True
```

Unless this is intentional, we could fix this by either explicitly checking for bools so this behaviour does not occur, or changing the `isinstance` check.

I've created a pull request which adds an extra check for bools, but feel free to suggest another fix!

pdtxie:
Fixed #622

**Changes proposed:**
Added an explicit check for booleans to prevent them from being converted to integers. The details of the issue are linked above.
```py
if elements and all(isinstance(item, bool) for item in elements):
    # Explicit check for bool, so it is not converted to int
    # as Python bools will have isinstance(.., int) as True
    return np.array(elements, dtype=bool)  
```

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -20,6 +20,8 @@ def test_make_array():
     assert test3.dtype == "<U3"
     test4 = ds.make_array(list(range(10)))
     assert test4.dtype == "object"
+    test5 = ds.make_array(True, False)
+    assert test5.dtype == "bool"
 
 
 def test_percentile():


adnanhemani:
Thanks for suggesting this fix - I just read the linked GH issue and think I understand the problem here. For more information, look at this SO: https://stackoverflow.com/questions/37888620/comparing-boolean-and-int-using-isinstance

Although this way to solve the issue is reasonable, I'd recommend to do some of the following instead:
* [Recommended] Put a check on the original line checking if `item` is an `int` that also checks `... and not isinstance(item, bool)`
* `type(item) == int`. Less recommended due to intricacies of `int` vs `np.integer` and checking types via equals isn't the recommended Pythonic way of doing things - but does keep the code cleaner, in a way.

Let me know your thoughts.

Please do also add a few, explicit unit tests that show this change as well.

adnanhemani:
Apologies, have been busy. For some odd reason, it won't let me run the Approval workflows here now?

Can you please close this PR and re-open it (mention this PR in it so we don't lose the context) and hopefully that will let me run the Approval workflows.

No tests should be failing AFAIK but we can tell that when the Approval Workflows run.

PR #624, opened 2025-04-02
Fix make_array() to correctly handle boolean arrays

GitGautamHub:
## Summary
- Fixes issue where `make_array(True, False)` was returning `[1, 0]` instead of `[True, False]`
- Explicitly checks for bools before converting to `int64`
- Adds `dtype=bool` when all elements are boolean

## Example Fix
```python
from datascience import make_array
print(make_array(True, False))  # Expected: array([ True, False], dtype=bool)

### DIFF
--- a/datascience/util.py
+++ b/datascience/util.py
@@ -23,25 +23,10 @@ def make_array(*elements):
     A simple way to make an array with a few elements.
 
     As with any array, all arguments should have the same type.
-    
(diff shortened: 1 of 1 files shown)


adnanhemani:
Will not be reviewing PRs that take out docstring and/or tests without explicit reasoning why.

Issue #618, opened 2024-10-29
Data type of row object items 


jedwin3210:
Current course staff from Data 8 here,

We were wondering if returning **"numpy.str_"** was an intentional feature for type(tbl.row(...).item(...)). I bring this up since type(tbl.column(...).item(...)) returns a str.  This means that in our test cases for jupyter notebooks, we should consider both of these types, but it would be beneficial to have both return a str for consistency and student readability. 

<img width="372" alt="Screenshot 2024-10-28 at 5 42 37 PM" src="https://github.com/user-attachments/assets/4c0218fb-0d06-4471-b8f4-8ac42ee25be3">

papajohn:
I don't think this behavior was intentional.

davidwagner:
I'm open to a pull request to change/fix this.

SebastianOuslis:
I think the core of this issue comes from the fact that the type returned by the `row` function and the `column` function are different.

<img width="267" alt="Image" src="https://github.com/user-attachments/assets/8e9822ce-dbeb-4f3d-81e1-a11bb244e7c8" />

This extends to the mismatch of types for other fields as well like integers.

<img width="286" alt="Image" src="https://github.com/user-attachments/assets/7f11438d-9fcb-4424-b880-6a3d6d57e714" />

I think the right thing to do would be to have both `row` and `column` return back the same type though this would be a potentially bigger scope of change than expected from the name of this issue.

adnanhemani:
@SebastianOuslis - not sure that ensuring that both Rows and Columns having the same data types representing them is possible and/or efficient.

The main thing to think about here is: columns should have elements of the same data type - rows (in 99% of cases) will be varied data types.

But all in all, I agree - discussing that today (unless there is absolutely no better option) is likely not productive.

PR #628, opened 2025-04-15
Update tables.py for Issue #618

ThomasL25:
Issue #618 states that the return values for the type of rows and columns result in different things. I believe this should keep it uniform between them.
(First time contributor)

**Changes proposed:**
Updated the Row.item() function to align better with the column output. 
Completed to solve [Issue #618 ](https://github.com/data-8/datascience/issues/618)

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -5709,7 +5709,8 @@ class Table(collections.abc.MutableMapping):
                 index = index_or_label
             else:
                 index = self._table.column_index(index_or_label)
-            return self[index]
+	    value = self[index]
+    	    return value.item() if hasattr(value, 'item') else value
 
         def __repr__(self):
             return 'Row({})'.format(', '.join('{}={}'.format(


adnanhemani:
While this is clever, these are the two things I'd like to see before I'm willing to approve:

1. Proper compilation: Approval Workflow build is failing due to indentation issues.
2. Tests that show the change being made.

adnanhemani (changes requested):
Please push the tests to re-run the CI/CD here. Seems like the AI in #639 did the same thing and the CI/CD worked there, so I think this will work here as well.

datascience/tables.py
```diff
@@ -5709,7 +5709,8 @@ def item(self, index_or_label):
                 index = index_or_label
             else:
                 index = self._table.column_index(index_or_label)
-            return self[index]
+	    value = self[index]
```
davidwagner, line 5712: Our automated tests complain that this line mixes tabs and spaces

PR #615, opened 2024-10-03
Add two unit tests for join method

vvedenskaya:
Added two unit tests to improve the test coverage of the join method in test_tables.py

PR #630, opened 2025-04-21
Add a unit test for join method coverage

joshu4lim:
[O] Wrote test for feature

**Changes proposed:**
Hello, I'm a student and a first-time contributor. 
One of the coverage missing in 'tables.py' is in the '_join' internal method, in lines [2039-2040](https://github.com/data-8/datascience/blob/master/datascience/tables.py#L2039-L2040) where it checks for an empty table. However, the 'join' method for external interface already has these checks in lines [2024-2025](https://github.com/data-8/datascience/blob/master/datascience/tables.py#L2024-L2025), preventing those lines from being covered. This makes it difficult to test '_join' by just calling 'join' as discussed in this [PR](https://github.com/data-8/datascience/pull/615). 

In this unit test, I decided to modify the 'join' method to skip those checks and allow '_join' to be covered. This way, the functionality of _join can be tested without directly invoking it.

Please let me know if any changes are needed to improve this implementation.

### DIFF
--- a/datascience/tables.py
+++ b/datascience/tables.py
@@ -2036,11 +2036,6 @@ class Table(collections.abc.MutableMapping):
 
     def _join(self, column_label, other, other_label=[]):
         """joins when COLUMN_LABEL is a string"""
-        if self.num_rows == 0 or other.num_rows == 0:
-            return None
-        if not other_label:
-            other_label = column_label
-
         self_rows = self.index_by(column_label)
         other_rows = other.index_by(other_label)
         return self._join_helper([column_label], self_rows, other, [other_label], other_rows)
--- a/tests/test_tables.py
+++ b/tests/test_tables.py
@@ -1660,18 +1660,6 @@ def test_join_without_other_label(table, table2):
     2      | c      | 3     | two
     """)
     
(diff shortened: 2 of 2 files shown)


adnanhemani:
I'm not really in favor of testing in such roundabout ways. If we have dead code, I would much rather remove it instead.

adnanhemani (changes requested):
I'm not really in favor of testing in such roundabout ways. If we have dead code, I would much rather remove it instead.

joshu4lim:
Removed the proposed roundabout test case and removed the dead code inside the _join internal method on lines [2039-2040](https://github.com/data-8/datascience/blob/master/datascience/tables.py#L2039-L2040). Also removed lines [2041-2042](https://github.com/data-8/datascience/blob/master/datascience/tables.py#L2041-L2042) because it would also be dead code based on the same reasoning for removing L2039-2040.

davidwagner (approved):

davidwagner:
@adnanhemani I'll let you comment, but it looks reasonable to me.

adnanhemani (approved):
LGTM - thanks for this change!

PR #635, opened 2025-05-17
Fix contradicting argument name in .sample() documentation (n -> k)

Omarr-kh:
[ ] Wrote test for feature

**Changes proposed:**
This PR fixes a documentation inconsistency in the .sample() method. The argument name in the documentation was listed as n, which contradicts the actual function signature and docstring that use k.


davidwagner (approved):

davidwagner:
Looks reasonable to me, but I'd like to wait until we get the continuous integration fixed so we can be sure the tests are passing, before I merge this.  Sorry about the delay, and thank you for the pull request.

docs/reference-nb/datascience-reference.ipynb
```diff
@@ -6488,4 +6488,4 @@
  },
  "nbformat": 4,
  "nbformat_minor": 2
-}
```
adnanhemani, line 6491: nit: Ideally we should revert this change.

adnanhemani:
@Omarr-kh can you please fix the nit and repush? I believe the CI/CD should be fixed now!

Omarr-kh:
@adnanhemani It should be fixed now.

PR #650, opened 2025-08-27
Fix regex warning

River739:
[ ] Wrote test for feature

**Changes proposed:**
<
1)  Fixed invalid regex escape sequence warning in "tests/test_formats.py" by using a raw string.
2)  Made "test_date_formatter_format_value" portable by skipping the test on platforms where "time.tzset()" is not available (e.g., Windows).
>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ docs/_build/
 docs/_autosummary/
 docs/_images/*.png
 docs/normal_data.csv
+venv/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -68,3 +68,4 @@ docs/_build/
 docs/_autosummary/
 docs/_images/*.png
 docs/normal_data.csv
+venv/
```
adnanhemani, line 71: Would prefer to not put this change within this PR.

adnanhemani (approved):
Overall LGTM

PR #652, opened 2025-09-04
Fix ipython directive error by adding blank lines after directives in tutorial.rst (GH#648)

Aniketsy:
This PR fixes the Sphinx error:
"maximum 4 argument(s) allowed, 17 supplied"
by ensuring a blank line is present after each `.. ipython:: python` directive in tutorial.rst.

PR #655, opened 2025-10-07
Fix Docs and Update URLs for Tutorials Notebook

adnanhemani:
[N/A] Wrote test for feature

**Changes proposed:**
Due to the merging of #652, we found a new bug that causes the `make docs` to completely error rather than the previous behavior, which was to ignore the last Python block in `tutorial.rst`. This change fixes this.

Additionally, I've updated the URLs that are referring to test datasets, which were moved earlier in the Textbook GH repo. Changing these URLs will allow users to re-run this notebook manually again on their local machines.


pancakereport (approved):
thanks for the fix!

PR #661, opened 2025-10-29
feat(util): add seed parameter for reproducible sampling 

jedwin3210:
- [x] Wrote test for feature
- [x] Added changes to CHANGELOG.md
- [x] Bumped version number (delete if unneeded)

**Changes proposed:**

- Add optional seed parameter to sample_proportions() and proportions_from_distribution()

- Fix column_name parameter bug in proportions_from_distribution()


sean-morris:
@jedwin3210 Thanks for rolling this so quickly. It looks great to me. I added @pancakereport as a reviewer

pancakereport (approved):
Thanks Edwin, looks good. 

My quick poking around the code shows that this is the only instance of `np.random.default_rng()`

pancakereport:
Actually - I think we should bump the version number.

pancakereport (approved):

PR #670, opened 2025-12-07
Implemented code coverage test case for _as_label function

MossIV:
[X ] Wrote test for feature

**Changes proposed:**
Test case tests the _as_label function to see if it raises a ValueError exception along with the appropriate exception message


davidwagner:
We generally prefer to test the externally visible API rather than internal private methods.

MossIV:
Got it, i'll keep that in mind for the future. Thanks for letting me know

PR #673, opened 2025-12-18
Add CONTRIBUTING, CI, and fix maps icons/tiles

Nishant141:
Adds CONTRIBUTING.md and CI workflow. Fixes folium tile attribution error for string tiles and normalizes BeautifyIcon option keys so icons expose `textColor`. Runs tests locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,31 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


davidwagner:
Thanks for all the suggested changes!  Fixes to tiles issues seems valuable.  Generally I'd prefer to have logically different changes split into separate pull requests: e.g., one for changes to workflow, one for changes to documentation about contributing, one for maps/tiles.   Can you split this up? I don't understand the benefit of this new workflow file over what we already have in .github/workflows/run_tests.yml; can you help me understand that?  Also isn't most everything in CONTRIBUTING already covered in DEVELOPERS.md -- is there value to that?

